PR #2, opened 2026-03-10
Add Google Calendar and Google Meet integration

Emilianouz:
This PR introduces Google Calendar integration to automatically create
meeting events and generate Google Meet links when a booking is created.

Includes:
- OAuth authentication with project Google account
- Added Google Calendar API service to create events and generate Google Meet links.
- Created API endpoint for meeting creation with necessary validation.
- Documentation updates for setup and testing


SaraTahir28:
This looks great, the view is clean and easy to follow. Request validation is handled properly and the Google Calendar logic is separated into the google_calendar_service, which keeps the view simple.

Error handling for invalid JSON and missing fields is also good.

One small improvement could be removing the unused render import but I know later we will need it when we integrate it wih frontend so lets keep it.

The use of csrf_exempt is fine for the prototype but should be reconsidered when authentication is added later on.

Overall the structure looks solid and the API responses are well designed for views.py.

SaraTahir28:
In google_calender_service.py

A docstring explaining what this module does will be beneficial e.g something like this. You can paste this in the file actually. 

This module is the core integration layer between Django backend and the Google Calendar API.
It handles:
1. loading OAuth credentials
2. refreshing tokens
3. building the Google Calendar service client
4. constructing the event payload
5. creating the event
6. generating a Google Meet link
7. returning the event ID + Meet link

SaraTahir28:
Makes sense. I will stop adding the labels. It was done out of habit. 😂

PR #43, opened 2026-03-24
Update /user/ endpoint and migration guidance

SaraTahir28:
**User endpoint update**
Added role and status fields to the /user/ endpoint so the dashboard can redirect users based on their role and account status.
This ensures that the frontend has all the necessary information to handle user navigation and permissions correctly.
**Database migration guidance**
If someone pulls main and encounters migration issues (e.g., missing tables or conflicts), reset migrations locally by deleting old migration files and the SQLite database, then recreate migrations:
**Setup**
Run Migrations (if needed after merging main)

1.cd Backend
If you encounter migration conflicts
`rm db.sqlite3`
`find . -path "*/migrations/*.py" -not -name "__init__.py" -delete`
`python manage.py makemigrations`
`python manage.py migrate`

2.Start backend:

python manage.py runserver

3.Start frontend:

cd ../Frontend
npm run dev

Open:
http://localhost:5173/

Expected behaviour

1. When you Click “Continue with Google”
2. you will be redirected to Complete Google login
3. After login - redirected back to the app
4. The booking UI (calendar + timeslots) should now be visible.

### DIFF
--- a/Backend/README.md
+++ b/Backend/README.md
@@ -155,20 +155,13 @@ GOOGLE_CLIENT_SECRET=your_client_secret
 
 ⚠️ Never commit this file
 
-3. Run migrations & server
-**Database Migrations After Merging `main`**
-
-If you pull the `main` branch and encounter migration issues (e.g., missing tables or conflicts), reset migrations locally by deleting old migration files and the SQLite database, then recreate migrations:
-
-```bash
-rm db.sqlite3
-find . -path "*/migrations/*.py" -not -name "__init__.py" -delete
-python manage.py makemigrations
+3. Run migrations & start the server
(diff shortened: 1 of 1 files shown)


duncanparkinson (changes requested):
Good work on a significantly smaller PR!

Unfortunately I still think this needs to be split further. I know it's already small, but it's about how many independent changes each PR makes, and in this case the PR can't be merged yet because of issues with the conflicting migrations advice, which shouldn't really affect us being able to merge in the small user endpoint changes.

This is a definite improvement, so just remember to think about whether using a word like "and" in the PR title is an indication that it might be doing too much and I think things will be excellent ☺️

PR #53, opened 2026-03-26
Logout Functionality with Google OAuth login Behavior.

### LINKED ISSUE #47: Create Logout Feature.
As a user, I want to be able to log out of the app from any page so that my session ends and I am redirected to the login page. The feature should clear the session on the backend and update the frontend state so that the app renders the login page after logout.
**Frontend**

1. Create a reusable LogoutButton component (wraps existing ActionBtn)
2. Create a Navbar component and import LogoutButton
3. Import Navbar in App.jsx so it appears on all pages
4. Ensure useAuth() context provides logout() function
5. Test logout from multiple pages
6. Connect it to backend to ensure we are actually clearing the session and logging out

SaraTahir28:
closes #47 
This PR:
1. Adds a logout button to the navbar for users to easily log out of the app.
2. Updates the app’s logout flow to immediately clear the user state on the frontend.
3. Ensures that the UI reflects logout instantly, instead of waiting for the backend response.
**Why this is done:**
Improves user experience by providing a clear logout option.
Avoids temporary display of protected content like booked metting/ agenda etc after logout.
**Setup**

 **Frontend**
`cd frontend`
`npm run dev`

 **Backend**
`cd backend`
`python manage.py runserver`

**Expected Behavior**
Logout button:

1. Click the logout button in the navbar.
2. The app UI immediately removes user info and redirects to the login page.
3. Django session is cleared in the backend.

**Google login flow**

1. Click “Login with Google.”
2. If multiple Google accounts are signed in, the user sees an account chooser.
3. If only one Google account is active, the user may be logged in automatically, this is standard OAuth behavior.
4. Closing and reopening the browser does not log the user out of Google; an active Google session persists.
5. No protected content should be visible after logout.
**Questions**
Do we actually need a logout button? Currently, logging out only clears the app session but does not log the user out of Google, so users can immediately log back in without entering credentials.
If the goal is security, should logout also fully log the user out of their Google account, or is clearing the app session enough?

### DIFF
--- a/Frontend/src/ProtectedApp.jsx
+++ b/Frontend/src/ProtectedApp.jsx
@@ -1,63 +1,59 @@
-import { useEffect, useState } from "react"; 
-import App from "./App"; // 
-import Login from "./components/pages/Login"; 
+import { useEffect, useState } from "react";
+import App from "./App";
+import Login from "./components/pages/Login";
 import { AuthProvider } from "./AuthContext";
-
-
 
 // ProtectedApp: wraps the main App and handles authentication checks
 export default function ProtectedApp() {
-	// `user` state will hold the current user info:
-	// - undefined: we are still checking if the user is logged in
-	// - null: user is not logged in
-	// - object: user is logged in (contains email, name, etc.)
-	const [user, setUser] = useState(undefined);
+  // `user` state will hold the current user info:
+  // - undefined: we are still checking if the user is logged in
+  // - null: user is not logged in
+  // - object: user is logged in (contains email, name, etc.)
+  const [user, setUser] = useState(undefined);
 
-	// useEffect runs once after component mounts
-	// It checks the backend to see if there is an active user session
-	useEffect(() => {
-		fetch("http://localhost:8000/auth/user/", {
-			credentials: "include", // ensures cookies (session) are sent with the request
-		})
-			.then((res) => {
-				if (res.status === 401) return null; 
-				return res.json(); 
-			})
-			.then((data) => {
-				setUser(data); // store the user object (or null if not logged in)
-			})
-			.catch(() => setUser(null)); // if fetch fails, treat as not logged in
-	}, []); // empty dependency array = runs only once on mount
-
-const API_BASE_URL = import.meta.env.VITE_API_BASE_URL;
-
-const [logoutError, setLogoutError] = useState(null);
-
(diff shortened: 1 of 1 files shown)


Frontend/src/ProtectedApp.jsx
```diff
@@ -17,21 +20,36 @@ export default function ProtectedApp() {
...
+				if (res.status === 401) return null; 
+				return res.json(); 
 			})
 			.then((data) => {
 				setUser(data); // store the user object (or null if not logged in)
 			})
 			.catch(() => setUser(null)); // if fetch fails, treat as not logged in
 	}, []); // empty dependency array = runs only once on mount
 
+function handleLogout() {
```
duncanparkinson: Can we fix this indentation please? It's not immediately clear that this function is correctly in the ProtectedApp scope because it isn't indented correctly.

Frontend/src/ProtectedApp.jsx
```diff
@@ -17,21 +20,37 @@ export default function ProtectedApp() {
...
 			.then((data) => {
 				setUser(data); // store the user object (or null if not logged in)
 			})
 			.catch(() => setUser(null)); // if fetch fails, treat as not logged in
 	}, []); // empty dependency array = runs only once on mount
 
+const API_BASE_URL = import.meta.env.VITE_API_BASE_URL;
+
+function handleLogout() {
+  setUser(null);
```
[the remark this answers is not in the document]
SaraTahir28: This is interesting because for me the behavior is different, Once I logout I always see the login screen only. even if I refresh it stays on login screen and clicking continue with google always takes me to the landing page
https://www.loom.com/share/7b9ac77298b946a6a6131748e1987c07.

Frontend/src/ProtectedApp.jsx
SaraTahir28: done

duncanparkinson (changes requested):
Thanks for the work on these changes! I think you've changed too much now though. My feedback was suggesting fairly minimal changes (just changing the `console.log` to an `alert` for now, as a placeholder we can fix later) and by changing more I think you've inadvertently introduced more issues.

Is there more to the decision to add logout error as state that I'm missing? If not I think the best thing to do would actually be to revert the last two commits and just make a very minimal change to create an alert on logout failure.

PR #54, opened 2026-03-26
Add a 404 page to improve routing feedback

katarzynakaz:
User Story:
As a user, I want to see clear feedback when I visit a page that does not exist, instead of the default React Router error screen.

Changes: 
Created a NotFound component with brand-consistent styling.

Added a wildcard path (*) to the router to ensure all not valid URLs are handled.

Added errorElement in the main router to catch crashes.

Why this matters:
Instead of seeing the "Unexpected Application Error" created by React, users now see "Take me home" button.

### DIFF
--- a/Frontend/src/App.jsx
+++ b/Frontend/src/App.jsx
@@ -33,14 +33,14 @@ const router = createBrowserRouter([
 ]);
 
 function App() {
-  return (
-    <div className="min-h-screen flex flex-col">
-      </ />
-      <div className="flex-1">
-        <RouterProvider router={router} />
-      </div>
-    </div>
-  );
+	return (
+		<div className="min-h-screen flex flex-col">
+			<Navbar />
+			<div className="flex-1">
+				<RouterProvider router={router} />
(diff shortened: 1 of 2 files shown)


Frontend/src/App.jsx
```diff
@@ -21,18 +21,21 @@ const router = createBrowserRouter([
...
+		path: "*",
+		Component: NotFound,
+	},
 ]);
 
 function App() {
   return (
     <div className="min-h-screen flex flex-col">
-      <Navbar />
+      </ />
```
duncanparkinson: Is this change intentional? It's stopping the app from building

Frontend/src/components/pages/NotFound.jsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { useNavigate } from "react-router-dom";
```
duncanparkinson: This import isn't needed anymore since we're using a standard `a` tag

PR #56, opened 2026-03-28
Volunteer Onboarding - Availability Slots Management

katarzynakaz:
Description
Implements the onboarding flow for volunteers to set up their initial availability.

VolunteerAvailabilityForm: Logic for adding recurring or specific date slots.

AddingSlotsBasket: Component for managing (adding/removing) the temporary list before final submission.

User Story and Acceptance Criteria
As a Volunteer, I can add and remove availability slots before saving to ensure my schedule is correct.

Onboarding Flow: Show the form when no slots exist; transition to the dashboard once saved.

UI: Single-row layout with DD-MM-YYYY formatting and HH:MM time.

To Do
Update the object to match the new DB design (including end_date, display initail date on recurring events to volunteer and end update start date calculations)


Frontend/src/index.css
```diff
@@ -263,4 +263,40 @@
...
+	}
+
+	.form-input-group-row {
+		@apply mb-6 flex flex-row items-center gap-3 p-3;
+	}
+
+	.form-input-group-row:hover {
+		@apply border-border-color;
+	}
+	ąć .form-label {
```
duncanparkinson: Some stray characters at the start of this line that will break this selector

Frontend/src/components/pages/VolunteerDash.jsx
```diff
@@ -95,25 +119,50 @@ const VolunteerDash = () => {
...
+						<p className="">
+							Let's start by selecting your availability for 1:1 sessions.
+						</p>
+
+						<VolunteerAvailabilityForm
+							volunteerId={activeVolunteer.id}
+							mode="edit"
+							whenFormSubmit={volunteerSubmitedFormWithSlots}
+							addedSlots={temporaryAddedSlotsStorage}
+							removeSlot={removeSlotFromTemporaryStorage}
```
duncanparkinson: Is there a reason these props need to live in VolunteerDash and not in VolunteerAvailablityForm? If we can move them down to VolunteerAvailablityForm we don't need to pass them here and all of the associated logic moves down too

duncanparkinson (changes requested):
I like this design with the list of entered slot rules, that's a neat solution. I don't think we need the day of the week selector, let's just use the start date to determine the day.

I've made notes for a few changes, but otherwise this is very good!

PR #57, opened 2026-03-28
Feature/ Volunteer Booking/ Pre-fill Trainee Data & Lock Email Field

katarzynakaz:
This changes the booking form by pulling trainee details directly from the Google Login.

Changes:

Auto-population: The Name and Email fields in the BookingForm are now automatically filled with the trainee's information.

Email is readOnly: The email field is now disabled. This ensures the booking is strictly linked to the logged-in account.

Flexible Name: The Name field remains editable. This allows users to change a Google name (in case it is a nickname or shorter version) before confirming the booking.

Visual Feedback: Added greyed-out styling to the email input to clearly communicate that it cannot be changed.

To test:
Open the booking flow.
Select any date/time.
Verify that your name and email are already there.
Try to click/edit the email (should be blocked).
Edit the name and click "Book meeting".


duncanparkinson (changes requested):
Do we actually need that intermediate page at all now? If we have the logged in user then we just need a "Book" button and we have all the user information we need already from the user model for the logged in user (i.e., we pass nothing about the user in the form since we don't let people book for anyone other than themselves).

I really like the thought of letting users change their name to be different from their Google account. However we probably don't want to make them enter it every time? Maybe this should be a "nice to have" and something that could be changed on a user profile page instead?

duncanparkinson (approved):

PR #58, opened 2026-03-29
Database Models for SlotsRule and Booking

### LINKED ISSUE #39: Create Models for Database, Bookings and Slots
Create Models in Database 

- Bookings
-Slots 
- Relationship between users, booking and slots.

Emilianouz:
### **Models for SlotsRule and Booking**
Endpoint for current user.

## **User:**
- Authentication is handled with Google OAuth.
- We keep AbstractUser so Django admin/auth still works normally.
- `username` still exists because it comes from AbstractUser.
- ‘group’ is a nice feature.

## **SlotRule Model:**   

It's the volunteer availability rule

Field: Start_time (The field has multiple purpose)
 For regular rules:
- the first datetime anchor, it will by used the hour (time-of-day) 

For non-regular rules:
- the exact datetime of the one-off slot

Field:  repeat_until (The field has multiple purpose)
 For regular rules:
- If not null indicates regular rule.
- It’s the end date of the rule
- permits calculate how many times the rule will be repeated

For non-regular:
- null

 Classes Meta permit add configuration to the data fields, nested in the model.

## **Booking Model:**
Concrete scheduled session between one trainee and one volunteer.
This is an actual booked occurrence, not an availability rule.

      Run after updating:
      
      python manage.py migrate
      

# **ENDPOINT:**
To identify the current user ( no need to know user ID), also can update their info.

`/api/me/ `so users update themselves safely

GET   `/api/me/`		returns current logged-in user
PATCH `/api/me/`		updates ONLY that user

**How to test it:**

In the browser go to:
`http://localhost:8000/api/me/`
GET shows the information of the current user.
PATCH: Change information, example GROUP in the form and press PUT

How to test it with Curl 
need to find the sessionid and csrftoken:

In Chrome:

Open DevTools
Go to Application
Open Cookies and copy:
_sessionid_
_csrftoken_

Then replace those real values in curl (capital).

    curl -X PATCH http://localhost:8000/api/me/ \
      -H "Content-Type: application/json" \
      -H "X-CSRFToken: _CSRF_TOKEN_" \
      -H "Cookie: sessionid=_SESSION_ID_; csrftoken=_CSRF_TOKEN_" \
      -d '{"role":"volunteer"}'

And to test GET:

    curl http://localhost:8000/api/me/ \
      -H "Cookie: sessionid=_SESSION_ID_; csrftoken=_CSRF_TOKEN_"

Closes #39


Backend/core/models.py
```diff
@@ -15,10 +16,108 @@ class User(AbstractUser):
...
+            errors["volunteer"] = "Selected volunteer must have role 'volunteer'."
+
+        if self.trainee_id and self.volunteer_id and self.trainee_id == self.volunteer_id:
+            errors["volunteer"] = "A user cannot book a session with themselves."
+
+        if errors:
+            raise ValidationError(errors)
+
+    def __str__(self):
+        return f"session {self.date} | {self.trainee} and {self.volunteer}"
```
duncanparkinson: `self.date` should be `self.start_time` I think?

duncanparkinson (changes requested):
Good work! A few things need addressed which I've left comments about.

One other thing: we need the migration committed to this PR so the rest of the team don't need to generate the migration themselves.

duncanparkinson:
> All requested changes have been addressed: I've removed the redundant fields and included the necessary migrations. Thanks for the feedback @duncanparkinson

Thanks for that. We do still need the email declaration for the user table though, just not with the length reduction. It was correct that you had a unique constraint on email: we can't have multiple users with the same email address.

PR #69, opened 2026-04-01
API - SlotRule creation endpoint 

Emilianouz:
This PR introduces the ability to create SlotRule records via the API. POST /api/slot-rules/

- Added SlotRuleSerializer for validation and creation
- Implemented POST /api/slot-rules/ endpoint
- Added validation: volunteer must have role "volunteer", repeat_until must not be before start_time

**SlotRuleSerializer**

Serializer for creating SlotRule records.
We accept volunteer by user ID.
>     Example:
>     {
>         "volunteer": 3,
>         "start_time": "2026-03-20T15:00:00Z",
>         "repeat_until": "2026-04-20",
>         "group": "all"
>     }
> 
slot_rule.full_clean() runs model-level validation.
        

### Important Notes:
volunteer is the user ID
repeat_until = null means one-off slot
repeat_until with a date means recurring rule
group is optional

**Use**

Example POST body

```
{
  "volunteer": 3,
  "start_time": "2026-03-20T15:00:00Z",
  "repeat_until": "2026-04-20",
  "group": "all"
}
```

Example curl
```
curl -X POST http://localhost:8000/api/slot-rules/ \
  -H "Content-Type: application/json" \
  -d '{
    "volunteer": 3,
    "start_time": "2026-03-20T15:00:00Z",
    "repeat_until": "2026-04-20",
    "group": "itd"
  }'
```

Close https://github.com/SaraTahir28/Pair-Scheduling-APP/issues/70

### DIFF
--- a/Backend/core/serializers/slot_rule_serializer.py
+++ b/Backend/core/serializers/slot_rule_serializer.py
@@ -18,17 +18,18 @@ class SlotRuleSerializer(serializers.ModelSerializer):
             "volunteer"
         ]
 
-    def validate_date(self, attrs):
-        
+    def validate_repeat_until_is_after_start_time(self, attrs):
         start_time = attrs.get("start_time")
         repeat_until = attrs.get("repeat_until")
-
         if repeat_until is not None and start_time is not None:
             if repeat_until < start_time.date():
                 raise serializers.ValidationError(
                     {"repeat_until": "repeat_until cannot be earlier than start_time date."}
                 )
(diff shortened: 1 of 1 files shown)


duncanparkinson (changes requested):
I didn't catch this on my first time reading the code: we need to make `volunteer` read-only and inject the correct user directly so users can't create slots for other people! I've put more details on the comment I've added to the file ☺️

PR #81, opened 2026-04-01
Feature/csrf protection

Emilianouz:
This PR removes the previous `csrf_exempt` workaround and fully restores Django’s built‑in CSRF protection. 

The frontend now uses `axios` for all authenticated requests, replacing the earlier fetch implementation. We introduced a CSRF helper and `axios` interceptor to ensure session cookies and CSRF tokens are consistently sent.

On the backend, the meeting creation endpoint has been upgraded to a DRF APIView with IsAuthenticated, ensuring only logged‑in users can create meetings and that CSRF + session authentication are correctly enforced.

Coauthor @Emilianouz @SaraTahir28


Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -5,6 +5,8 @@ import Calendar from "../groups/Calendar";
 import TimeSlotGroup from "../groups/TimeSlotGroup";
 import BookingForm from "../groups/BookingForm";
 import { volunteersDetails } from "../../data/UserData";
+import api from "../../api/axiosclient";
+
```
duncanparkinson: The file is `axiosClient.js`, not `axiosclient.js`. This will work on a Mac but Linux is case sensitive for this so it will break when we try to actually put it on a server. Also an extra blank line we don't need under this import (just a nitpick, but trying to clean things up a little)

PR #86, opened 2026-04-04
Feat/available slots calculation

### LINKED ISSUE #35: Database for Volunteers Date and time slots Record
- As a trainee I want to be able to view all volunteers available time slots
- We have offered time slot rules on our volunteer form . when they submit that goes to our database through our backend( learn about layers) 
-  Backend querys the database for available date and timeslots and recurring etc everything. 
- Trainee dashboard talks to our Backend and Backend queries the database and all timeslots gets dispalyed on the trainee calendar dashboard

Emilianouz:
Available Slots feature that generates bookable time slots for volunteers based on SlotRule. It includes a service layer, API endpoint, and test.

### Added extensive tests for both service and API:

**Service tests**

Single and multiple one-off slots
Exclusion of past slots
Expansion of recurring rules
Filtering out past occurrences in recurring rules

**API tests**

Authentication required (403 if unauthenticated)
Successful slot retrieval
Filtering by volunteer_id

**Service layer:** build_available_slots

- Added AvailableSlot dataclass to represent computed slots.
- Implemented build_available_slots

**API endpoint:**  /available-slots/

- New authenticated endpoint via AvailableSlotsView.
- Returns a list of available slots derived from SlotRule

**URL configuration**
Registered new route:
/available-slots/

### Improvements done:

View filtering:
- Filter available slots by requesting user group (including "all")
- Add query parameter filtering for group and role
- Exclude already booked slots by matching (volunteer_id, start_time). Find all existing bookings (from DB) that match those volunteers and start times vs build_available_slots, and then removes already booked.

Model:
- Fix SlotRule.occurrence_start_times() to correctly compare datetime.date with repeat_until

Tests:
- Add coverage for group-based visibility
- Add coverage for query parameter filtering (group, role)
- Add tests for time window filtering
- Add tests ensuring booked slots are excluded while other slots remain available

**Available slots are now filtered against existing bookings:**
We fetch all bookings matching (volunteer_id, start_time) pairs from generated slots
Already booked slots are excluded from the response

### Notes

We use query parameters to support filtering by both group and role, following the existing pattern:

`?volunteer_id=...`

This allows extending the same endpoint to support:

```
/available-slots/?group=itd
/available-slots/?role=trainee
/available-slots/?group=itd&role=admin
Group filtering: important distinction

```

**Updated rule business**
Any user can host sessions: a host (currently stored in the volunteer field) can be any user (trainee, volunteer, staff, etc.). Slots can be filtered based on the host’s role

The volunteer field represents the host of the session, not strictly a user with role "volunteer". Renaming this field (e.g. to host) could improve clarity, but is out of scope for this PR.

**Future work**

Add more query-based filters (e.g. language, tags, labels)
Consider renaming volunteer → host for clarity

Close #35


Emilianouz (commented):
The merge conflicts were solved.

duncanparkinson (changes requested):
Really good work on this!

A couple of small changes needed in the test setup in a couple of places, and then a bigger comment about where the duplicate booking logic lives. The logic is sound but the code would be even better with some tweaks I think ☺️

Emilianouz:
Changes:

- [x] Logic of filter out already booked slots was moved from AvailableSlotsView to build_available_slots() method.

- [x] Removed database view tests and tests related to service behaviour.

- [x] Updated make_user with group.

**Extra:**

Added available_slot_service_tests, moving build_available_slots() tests out of the view and introducing mocked Booking queries:
- Slot generation (one-off and recurring)
- Time filtering (past exclusion and limits)
- Mock Booking.objects.filter to test booking exclusion logic without DB
- Ensure only exact (volunteer_id, start_time) pairs are excluded

duncanparkinson (changes requested):
This is looking really good, and the checklist of completed items is very helpful when re-reviewing so thank you!

One final small change needed: can we still have the retrieval of Bookings from the DB happen in the view, and pass those in as an additional argument to `build_available_slots` please? We want to keep database logic out of our service class so it only handles data that is passed into it, and then the tests that use the database can all live in the view tests, and our service tests can deal with just the actual logic of that function.

duncanparkinson (approved):
Really great work on this! Do you think that the changes to the tests made the tests simpler? They definitely look cleaner to me like this now.

I've added a couple of very small things to fix but have also approved this for merging. You can merge this in once those things are tidied up ☺️

PR #94, opened 2026-04-09
Reafctor MeView to Current Profile View

SaraTahir28:
This PR refactors the MeView class to follow clearer naming practices by renaming it to CurrentProfileView. The new name better communicates the intent of the endpoint which represents the profile of the currently authenticated user.

Purpose of the Endpoint
 1. Since all users authenticate via Google, they will not be updating core identity fields such as email or username through this endpoint.
2. The endpoint exists to allow users to view their own profile and, in the future, update dynamic attributes such as:
 Group choices (e.g.ITP, SDC,launch for trainees)
3. Language or specialization tags (e.g., volunteers marking themselves as Python,JavaScript specialists)
4. These attributes are expected to evolve as users progress through courses or host sessions, making a dedicated self‑service profile endpoint valuable.

**Note:**
This will need to be approved before the testing PR for this endpoint and view (https://github.com/SaraTahir28/Pair-Scheduling-APP/pull/92) and then in testing PR we will need to update routing to api/profile from api/me.

### DIFF
--- a/Backend/core/tests/test_current_user_profile.py
+++ b/Backend/core/tests/test_current_user_profile.py
@@ -12,7 +12,7 @@ def test_current_user_profile_get_authenticated(client, django_user_model):
     )
     client.login(username="sara", password="securepass123")
 
-    response = client.get("/api/me/")
+    response = client.get("/api/profile/")
     assert response.status_code == 200
 
     data = response.json()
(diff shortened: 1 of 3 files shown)


duncanparkinson (changes requested):
Looks good but will need to be updated now that #92 has been merged

duncanparkinson (approved):

PR #104, opened 2026-04-10
Cypress tests for the Volunteer flow

katarzynakaz:
This PR is for the Cypress integration test for the Volunteer onboarding.

Changes
Cypress: Added tests for single/recurring slots, basket removal, and dashboard rendering.
Auth Integration: Switched VolunteerDash to useAuth().
Volunteer Flow: Added fallback profile image (duncan.png).

Testing
cd Frontend && npm run dev
npx cypress open
Run e2e and select Integration-Volunteer.cy.jsx:

### DIFF
--- a/Frontend/cypress/e2e/Integration-VolunteerOnboarding.cy.jsx
+++ b/Frontend/cypress/e2e/Integration-VolunteerOnboarding.cy.jsx
@@ -1,15 +1,16 @@
 describe("Volunteer flow", () => {
-	it("allows volunteer to save a single date availability", () => {
+	beforeEach(() => {
 		cy.intercept("GET", "**/auth/user/", {
 			statusCode: 200,
 			body: {
 				id: 1,
 				name: "Duncan Parkinson",
-				email: "duncan@test.com",
-				is_volunteer: true,
+				email: "duncan@example.com",
 			},
 		}).as("getUser");
+	});
 
+	it("allows volunteer to save a single date availability", () => {
 		cy.visit("/volunteer-dash");
 		cy.wait("@getUser");
 
@@ -35,16 +36,6 @@ describe("Volunteer flow", () => {
 	});
 
 	it("allows volunteer to save a recurring availability", () => {
-		cy.intercept("GET", "**/auth/user/", {
-			statusCode: 200,
-			body: {
-				id: 1,
-				name: "Duncan Parkinson",
-				email: "duncan@test.com",
-				is_volunteer: true,
-			},
-		}).as("getUser");
-
 		cy.visit("/volunteer-dash");
 		cy.wait("@getUser");
 
@@ -78,16 +69,6 @@ describe("Volunteer flow", () => {
 	});
 
 	it("allows volunteer to remove an item from the basket", () => {
-		cy.intercept("GET", "**/auth/user/", {
-			statusCode: 200,
-			body: {
-				id: 1,
-				name: "Duncan Parkinson",
-				email: "duncan@test.com",
-				is_volunteer: true,
-			},
-		}).as("getUser");
-
 		cy.visit("/volunteer-dash");
 		cy.wait("@getUser");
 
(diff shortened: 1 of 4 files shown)


Frontend/cypress/e2e/Integration-VolunteerOnboarding.cy.jsx
```diff
@@ -0,0 +1,137 @@
+describe("Volunteer flow", () => {
+	it("allows volunteer to save a single date availability", () => {
+		cy.intercept("GET", "**/auth/user/", {
+			statusCode: 200,
+			body: {
+				id: 1,
+				name: "Duncan Parkinson",
+				email: "duncan@test.com",
```
duncanparkinson: Can we always use `@example.com` for examples please. It's reserved for testing purposes so we can safely send emails to it without them ever being accidentally delivered to someone!

Frontend/cypress/e2e/Integration-VolunteerOnboarding.cy.jsx
```diff
@@ -0,0 +1,137 @@
...
+	it("allows volunteer to save a recurring availability", () => {
+		cy.intercept("GET", "**/auth/user/", {
+			statusCode: 200,
+			body: {
+				id: 1,
+				name: "Duncan Parkinson",
+				email: "duncan@test.com",
+				is_volunteer: true,
+			},
+		}).as("getUser");
```
duncanparkinson: Is this setup identical across all the tests? If so it could be cleaned up using a before hook:

```
beforeEach(() => {
    cy.intercept("GET", "**/auth/user/", {
        statusCode: 200,
        body: { id: 1, name: "Duncan Parkinson", email: "duncan@test.com", role: "volunteer" },
    }).as("getUser");
});
```

Frontend/cypress/e2e/Integration-VolunteerOnboarding.cy.jsx
```diff
@@ -0,0 +1,137 @@
+describe("Volunteer flow", () => {
+	it("allows volunteer to save a single date availability", () => {
+		cy.intercept("GET", "**/auth/user/", {
+			statusCode: 200,
+			body: {
+				id: 1,
+				name: "Duncan Parkinson",
+				email: "duncan@test.com",
+				is_volunteer: true,
```
duncanparkinson: I don't think this matches the way the User model is setup? I don't think we are using roles at all for this anymore, and I don't think we have an `is_volunteer` field, so I think this mock is potentially confusing

duncanparkinson (changes requested):
Good work on this!

The main question I have though is are these really e2e tests? They seem to be frontend integration tests. An e2e test would go all the way through the frontend to the backend and the database, and none of these tests do because they are mocking that interface with `cy.intercept`. Integration tests are still useful, but we also need a couple of true e2e tests in order to actually test that the FE and BE are working together.

katarzynakaz:
> Good work on this!
> 
> The main question I have though is are these really e2e tests? They seem to be frontend integration tests. An e2e test would go all the way through the frontend to the backend and the database, and none of these tests do because they are mocking that interface with `cy.intercept`. Integration tests are still useful, but we also need a couple of true e2e tests in order to actually test that the FE and BE are working together.

They’re frontend integration tests with mocks.
The goal here was to first cover the flows with stable tests, refactor and then use that as a foundation for adding the system tests.

Frontend/cypress/e2e/Integration-VolunteerOnboarding.cy.jsx
katarzynakaz: updated as requested

Frontend/cypress/e2e/Integration-VolunteerOnboarding.cy.jsx
katarzynakaz: added as requested

duncanparkinson (approved):

PR #105, opened 2026-04-10
Feature/update to production URL's

SaraTahir28:
This PR  Introduces environment‑specific configuration using CYF domain URLs for both frontend and backend, replacing hardcoded localhost values.

**Testing**  

Verified by exporting DJANGO_ENV=production locally and running `python manage.py runserver.` The console output confirmed .env.production was loaded and FRONTEND_URL resolved to the CYF domain. To reproduce:

1. Run `export DJANGO_ENV=production ` your backend virtual environment.
2. Start the server with `python manage.py runserver`
3. Check logs for DJANGO_ENV: production and FRONTEND_URL loaded: https://pairscheduler-frontend.hosting.codeyourfuture.io

Additional Notes  
During setup we faced an issue with our Google Cloud Console project which required enabling 2‑factor authentication. This was resolved by configuring the Google Authenticator app to secure project access.

### DIFF
--- a/Backend/booking_system/settings.py
+++ b/Backend/booking_system/settings.py
@@ -34,13 +34,13 @@ GOOGLE_OAUTH2_CLIENT_SECRET = env("GOOGLE_CLIENT_SECRET")
 
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = "django-insecure-(#1c#^ypty_2-2!3g^uu9p@z0=#2+7nv8qz#swf@yg&okj0ld1"
+SECRET_KEY = os.environ.get("DJANGO_SECRET_KEY")
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = DJANGO_ENV != "production"
 
 ALLOWED_HOSTS = [
-    "pairscheduler-backend.hosting.codeyourfuture.io",  # production backend
+    "pairscheduler-backend.hosting.codeyourfuture.io",
     "localhost",
     "127.0.0.1",
 ]
@@ -256,6 +256,3 @@ CSRF_TRUSTED_ORIGINS = [
     "http://127.0.0.1:5174",
     "https://pairscheduler-frontend.hosting.codeyourfuture.io",
 ]
-
-print("DJANGO_ENV:", DJANGO_ENV)
-print("FRONTEND_URL loaded:", FRONTEND_URL)
--- a/Backend/booking_system/settings.py
+++ b/Backend/booking_system/settings.py
@@ -34,7 +34,7 @@ GOOGLE_OAUTH2_CLIENT_SECRET = env("GOOGLE_CLIENT_SECRET")
 
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = os.environ.get("DJANGO_SECRET_KEY")
+SECRET_KEY = env("DJANGO_SECRET_KEY")
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = DJANGO_ENV != "production"
--- a/Backend/booking_system/settings.py
+++ b/Backend/booking_system/settings.py
@@ -18,27 +18,32 @@ import os
 BASE_DIR = Path(__file__).resolve().parent.parent
 # This tells Django to read  .env file so  Google credentials become available.
 env = environ.Env()
+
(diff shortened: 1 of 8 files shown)


Backend/booking_system/settings.py
```diff
@@ -18,27 +18,32 @@
...
 
-# Base URL of the backend API
-# Used to build redirect URIs for OAuth2 login flow
-BASE_API_URL = "http://localhost:8000"
-
-# Quick-start development settings - unsuitable for production
-# See https://docs.djangoproject.com/en/6.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = "django-insecure-(#1c#^ypty_2-2!3g^uu9p@z0=#2+7nv8qz#swf@yg&okj0ld1"
```
duncanparkinson: Now that we're prepping to put this in production the secret key needs to be moved out to .env (and it **must** be a different key from this one, because this key has been pushed to GitHub)

Backend/booking_system/settings.py
```diff
@@ -248,5 +254,8 @@
     "http://127.0.0.1:5173",
     "http://localhost:5174",
     "http://127.0.0.1:5174",
-    "http://localhost:3000",
+    "https://pairscheduler-frontend.hosting.codeyourfuture.io",
 ]
+
+print("DJANGO_ENV:", DJANGO_ENV)
+print("FRONTEND_URL loaded:", FRONTEND_URL)
```
duncanparkinson: We don't want this debug when we merge this to `main`

Backend/booking_system/settings.py
```diff
@@ -208,7 +214,7 @@
     "http://127.0.0.1:5173",
     "http://localhost:5174",
     "http://127.0.0.1:5174",
-    "http://localhost:3000",
+    "https://pairscheduler-frontend.hosting.codeyourfuture.io",  # production frontend
```
duncanparkinson, line 217: I don't think we need the comment on this line

duncanparkinson (changes requested):
Good work on this! A couple of minor changes and one small (but very important) one

Backend/booking_system/settings.py
SaraTahir28: removed

Backend/booking_system/settings.py
SaraTahir28: Done.

Backend/booking_system/settings.py
SaraTahir28: Removed

duncanparkinson:
The CI test is failing because the `SECRET_KEY` variable name doesn't match between the test config and this branch. This is where it's set up for the CI testing:

https://github.com/SaraTahir28/Pair-Scheduling-APP/blob/46f9b47f9c033e11a7c3d40b9007f4e3705c8d27/.github/workflows/backend.yml#L28

SaraTahir28:
Fixed it. Thankyou. :)

Backend/booking_system/settings.py
```diff
@@ -18,27 +18,32 @@
...
-# Base URL of the backend API
-# Used to build redirect URIs for OAuth2 login flow
-BASE_API_URL = "http://localhost:8000"
-
-# Quick-start development settings - unsuitable for production
-# See https://docs.djangoproject.com/en/6.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = "django-insecure-(#1c#^ypty_2-2!3g^uu9p@z0=#2+7nv8qz#swf@yg&okj0ld1"
+SECRET_KEY = os.environ.get("DJANGO_SECRET_KEY")
```
duncanparkinson: Sorry, one more small change! Can we use `env()` here like we do on lines 32:33 please? Both work but they behave slightly differently if the key isn't defined and it's better for them to be consistent.

duncanparkinson (approved):
One small change I didn't catch before. Ready to merge once that's done so I've approved the branch for merging ☺️

Backend/booking_system/settings.py
SaraTahir28: Just did that. I am not sure about one thing,  I rebased this branch to main and I tried creating a meeting and its rasing a 400 bad request. Its  something in process to be done or a bug that needs fixing ?

PR #106, opened 2026-04-10
chore/ Back button added to trainee booking flow

katarzynakaz:
Adds a reusable BackBtn component with a chevron.

The trainee can go back to the function in the booking form, which removes the selected time from the URL, returning the user to the time-slot selection view.


Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -93,8 +99,13 @@ const TraineeBookingFlow = () => {
 						</>
 					)}
 
+					{/* when user selects date and time we render booking form and hide the cal */}
+					{/* {selectedTime && !isBookingConfirmed && ( */}
```
duncanparkinson: One last commented out section to remove

Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -93,8 +99,13 @@ const TraineeBookingFlow = () => {
...
 					{selectedTime && !isConfirmationPage && (
 						<div className="timeslot-col trainee-timeslot-width">
+							<div style={{ marginBottom: "16px" }}>
```
duncanparkinson: I think this is the only place using an inline style instead of a CSS class. Can we make a class for this please?

duncanparkinson (changes requested):

Frontend/src/components/pages/TraineeBookingFlow.jsx
katarzynakaz: done

Frontend/src/components/pages/TraineeBookingFlow.jsx
katarzynakaz: comment removed

duncanparkinson (approved):

PR #107, opened 2026-04-10
fix: Calendar past-month display is disabled

katarzynakaz:
This PR prevents users from navigating back to past months when booking a session. 

It compares the first day of the viewed month against the first day of the current month.


Frontend/src/components/groups/Calendar.jsx
```diff
@@ -58,6 +58,9 @@ const Calendar = ({
 	const firstDay = new Date(year, month, 1);
 	const daysInMonth = new Date(year, month + 1, 0).getDate();
 	const numOfEmptyFieldsBeforeFirstDay = (firstDay.getDay() + 6) % 7;
+	const today = new Date();
+	const currentMonth = new Date(today.getFullYear(), today.getMonth(), 1);
+	const isPastDate = firstDay <= currentMonth;
```
duncanparkinson: I think this variable name could be clearer. Maybe just `isCurrentMonth` would be better since we aren't going to allow past months anyway?

duncanparkinson (changes requested):

Frontend/src/components/groups/Calendar.jsx
katarzynakaz: variable renamed

duncanparkinson (approved):

PR #108, opened 2026-04-11
style: update booking form and slots basket UI

katarzynakaz:
Styling updates for the booking form and slots basket so that the components are consistent with the rest of the styles.


Frontend/src/components/groups/AddingSlotsBasket.jsx
```diff
@@ -1,44 +1,55 @@
 import React from "react";
 import { ActionBtn } from "../elements/Button";
+import { X } from "lucide-react";
```
duncanparkinson, line 3: Are we actually using this import? It looks like everywhere is still using a standard 'x' character

duncanparkinson (changes requested):

Frontend/src/components/groups/AddingSlotsBasket.jsx
katarzynakaz: Thank you, I intended to use the icon but forgot to swap it. Issue fixed.

katarzynakaz:
I replaced the 'x' character with the <X /> icon and updated the styling. 
Cypress Integration-VolunteerOnboarding.cy.jsx has to be updated to pass.
cy.contains("✕").click() 
should be changed to:
cy.get(".basket-delete-btn").click()

duncanparkinson (changes requested):
Great, thanks!

For the tests: those changes should be included in this PR. We want to keep the tests in sync with the production code because the tests should be our indication of what the code is supposed to be doing, so I definitely wouldn't split tests changes like that into a separate PR.

Once those tests are fixed this is good to go ☺️

katarzynakaz:
> Great, thanks!
> 
> For the tests: those changes should be included in this PR. We want to keep the tests in sync with the production code because the tests should be our indication of what the code is supposed to be doing, so I definitely wouldn't split tests changes like that into a separate PR.
> 
> Once those tests are fixed this is good to go ☺️

Thank you very much, that's the test updated.

duncanparkinson (approved):

PR #116, opened 2026-04-12
Feature/testing slot rule create view

SaraTahir28:
closes #95 
This PR introduces tests for the SlotRuleCreateView to ensure correctness, validation, permissions, persistence, and edge case handling.

**Coverage**

1. test_create_slot_rule_success: verifies slot rule creation with valid data.
2. test_slot_rule_persisted_in_db: confirms slot rule is saved with correct volunteer and group.
3. test_repeat_until_before_start_time: rejects invalid date ranges.
4. test_missing_required_fields: enforces required fields.
5. test_repeat_until_none: ensures repeat_until is optional.
6. test_unauthenticated_user_cannot_create: blocks unauthenticated users from creating slot rules.
7. test_multiple_slot_rules_for_same_user: ensures multiple slot rules can be created and persisted for the same user.
8. test_volunteer_auto_assigned: confirms volunteer is auto-assigned to the authenticated user and random IDs are ignored.

**How to Run** 

1. Cd Backend
2. Run `pytest test_slot_rule_create_view.py`
3. All test should pass.


Backend/core/tests/test_slot_rule_create_view.py
```diff
@@ -0,0 +1,114 @@
...
+        api_client.force_authenticate(user=user)
+        for group in ["piscine", "sdc"]:
+            payload = {
+                "start_time": "2026-04-12T10:00:00Z",
+                "repeat_until": "2026-04-15",
+                "group": group,
+            }
+            response = api_client.post(self.get_url(), payload, format="json")
+            assert response.status_code == 201
+        assert SlotRule.objects.filter(volunteer=user).count() == 2
```
duncanparkinson, line 112: This test works for now (and in general I like using a loop like that to generate the two POSTs) but it's going to break when we add `UNIQUE` constraints on `start_time` in the db. For now the simplest solution is probably just to explicitly create two different SlotRules

duncanparkinson (changes requested):
One small test change needed but otherwise the tests look good!

There seems to be a lot of crossover with an already merged PR though, with things like the `SECRET_KEY` changes? For situations like this make sure to do `git rebase main` and then force push the result of that so we have a clean PR with only the relevant changes

Backend/core/tests/test_slot_rule_create_view.py
SaraTahir28: Done.

duncanparkinson (approved):

PR #117, opened 2026-04-12
Available slots endpoint for calendar frontend JSON response

Emilianouz:
This PR refactors the available slots endpoint to align with current agreement returning to frontend JSON response to render in the calendar.

**Changes included**

- Refactor available_slots including name and img.
- Add serializer for available slots frontend response.
- Refactor AvailableSlotsView with serialize API response
- Refactor service test including user name.
- Update view tests available slots including name and img

**API response**

The endpoint now returns items in this format:

> [
>   {
>     "start_time": "2026-01-01T09:00:00Z",
>     "end_time": "2026-01-01T10:00:00Z",
>     "volunteer_id": 1,
>     "slot_rule_id": 1,
>     "name": "Duncan Parkinson",
>     "img": "/public/placeholder.png"
>   }
> ]

## Endpoint

**Use:**

`GET /api/available-slots/`

**Frontend usage**

If the frontend Axios client has baseURL = http://localhost:8000, use:

```
api.get("/api/available-slots/")
```

With filters:

```
api.get("/available-slots/", {
  params: {
    group: "itd",
    role: "volunteer",
  },
});
```

Notes
The img is fixed    ` "img": "/public/placeholder.png" `
Available query parameters: volunteer_id, group, role. 
Already booked slots are excluded before the response is returned
Testing

**Backend tests cover:**
- one-off slot generation
- recurring slot expansion
- booking window filtering
- booked slot exclusion
- endpoint filtering by group, volunteer, and role

Close #96

### DIFF
--- a/Backend/core/serializers/available_slot_serializer.py
+++ b/Backend/core/serializers/available_slot_serializer.py
@@ -3,6 +3,7 @@ from rest_framework import serializers
 class AvailableSlotSerializer(serializers.Serializer):
     slot_rule_id = serializers.IntegerField()
     start_time = serializers.DateTimeField()
+    end_time = serializers.DateTimeField()
     volunteer_id = serializers.IntegerField()
     name = serializers.CharField()
     img = serializers.CharField()
--- a/Backend/core/serializers/available_slot_serializer.py
+++ b/Backend/core/serializers/available_slot_serializer.py
@@ -1,0 +1,8 @@
+from rest_framework import serializers
+
+class AvailableSlotSerializer(serializers.Serializer):
+    slot_rule_id = serializers.IntegerField()
+    start_time = serializers.DateTimeField()
+    volunteer_id = serializers.IntegerField()
+    name = serializers.CharField()
+    img = serializers.CharField()
--- a/Backend/core/services/available_slots.py
+++ b/Backend/core/services/available_slots.py
@@ -25,7 +25,7 @@ def build_available_slots(rules, beginning_of_booking_window):
                     group=rule.group,
                     start_time=start_time,
                     end_time=start_time + timedelta(hours=1),
-                    name=rule.volunteer.get_full_name(),
+                    name=rule.volunteer.get_full_name() or rule.volunteer.get_username(),
                     img="/public/placeholder.png",
                 )
             )
--- a/Backend/core/services/available_slots.py
+++ b/Backend/core/services/available_slots.py
@@ -25,7 +25,7 @@ def build_available_slots(rules, beginning_of_booking_window):
                     group=rule.group,
                     start_time=start_time,
                     end_time=start_time + timedelta(hours=1),
-                    name=rule.volunteer.get_full_name() or f"{rule.volunteer.first_name} {rule.volunteer.last_name}".strip() or rule.volunteer.username,
+                    name=rule.volunteer.get_full_name(),
                     img="/public/placeholder.png",
                 )
             )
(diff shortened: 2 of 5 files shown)


Backend/core/services/available_slots.py
```diff
@@ -23,6 +25,8 @@ def build_available_slots(rules, beginning_of_booking_window):
                     group=rule.group,
                     start_time=start_time,
                     end_time=start_time + timedelta(hours=1),
+                    name=rule.volunteer.get_full_name() or f"{rule.volunteer.first_name} {rule.volunteer.last_name}".strip() or rule.volunteer.username,
```
duncanparkinson: Django already does the equivalent of `f"{rule.volunteer.first_name} {rule.volunteer.last_name}".strip()` when you call `get_full_name()` I think? I don't think we need that second `or` at all, just `rule.volunteer.get_full_name() or rule.volunteer.username` should be sufficient

Backend/core/serializers/available_slot_serializer.py
```diff
@@ -0,0 +1,8 @@
+from rest_framework import serializers
+
+class AvailableSlotSerializer(serializers.Serializer):
+    slot_rule_id = serializers.IntegerField()
+    start_time = serializers.DateTimeField()
```
duncanparkinson, line 5: Can we add `end_time` back in here please?

https://codeyourfuture.slack.com/archives/C0AK3D9P1DX/p1776071912606609

Backend/core/tests/test_available_slots_service.py
```diff
@@ -3,19 +3,34 @@
...
-from core.models import SlotRule
+from core.models import SlotRule, User
 from core.services.available_slots import build_available_slots, exclude_booked_slots
 
 NOW = timezone.now()
 FUTURE = timezone.now() + timedelta(days=1)
 
-def make_slot_rule(volunteer_id=1, start_time=None, repeat_until=None, rule_id=1, group="itd"):
+def make_slot_rule(
+        volunteer_id=1,
```
duncanparkinson: Can we match this indentation to the rest of the params please?

duncanparkinson (changes requested):
Looks good, just a couple of small changes needed

katarzynakaz (commented):
Looks great from the frontend side! The JSON matches what we discussed for the Calendar flow, and including the end_time is helpful for moving the calculation to the backend.

Backend/core/services/available_slots.py
Emilianouz: That’s correct; I was just trying to be cautious.

Backend/core/serializers/available_slot_serializer.py
Emilianouz: Changes are now implemented.

Emilianouz:
The changes have been applied. @duncanparkinson

Backend/core/services/available_slots.py
duncanparkinson: Handling potentially incorrect data is always a good idea, it's just that the middle part would also have been empty if the first part was empty so couldn't ever do anything (I'm basing this assertion on the method docs [here](https://docs.djangoproject.com/en/6.0/ref/contrib/auth/#django.contrib.auth.models.User.get_full_name) by the way). If we are allowing first and last name to be blank though then we do still need the `or rule.volunteer.get_username()` part though, so I think you've overcorrected slightly here.

(Note that `get_username()` is recommended over directly accessing `username` [here](https://docs.djangoproject.com/en/6.0/ref/contrib/auth/#django.contrib.auth.models.User.get_username))

Emilianouz: Understood, changes pushed.

duncanparkinson (approved):

PR #120, opened 2026-04-13
Feat/avoid slot rule duplication

### LINKED ISSUE #119: Preventing Duplicates in SlotRules
Enforce uniqueness in the slot rules table to prevent volunteers from having identical rules for the same time slot.

Acceptance Criteria:

> A volunteer shouldn’t have two identical slot rules at the same time.

Emilianouz:
Enforce uniqueness in the slot rules table to prevent volunteers from having identical rules for the same time slot.

- Added a model constraint to prevent slot_rule duplication and included the associated migration.
- Add validation in serializer to prevent overlapping slot rulesfor volunteers

Run:
`
python manage.py migrate
`
to apply the migration.

Close #119


Backend/core/models.py
```diff
@@ -51,6 +51,12 @@ class SlotRule(models.Model):
 
     class Meta:
         ordering = ["volunteer_id", "start_time"]
+        constraints = [
+        models.UniqueConstraint(
+            fields=["volunteer", "start_time"],
+            name="unique_volunteer_start_time_slot_rule",
+        )
+    ]
```
duncanparkinson: This indentation is incorrect to the point I'm not sure the code will actually execute correctly? Lines 55:59 all need to be indented a level further. Please try to be extra careful with Python formatting because it is a language unusually prone to errors if things like indentation are wrong!

Backend/core/serializers/slot_rule_serializer.py
```diff
@@ -30,6 +30,17 @@ def validate_repeat_until_is_after_start_time(self, attrs):
 
     def validate(self, attrs):
         attrs = self.validate_repeat_until_is_after_start_time(attrs)
+        volunteer = self.instance.volunteer if self.instance else self.context["request"].user
```
duncanparkinson, line 33: No need to change anything, just a note to consider: if we ever directly instantiate this Serializer (like in a test or a script) this is likely to fail in ways that might not be clear. `context["request"]` would be None in that situation, and so `.user` couldn't be called on it.

Don't change anything for now because we aren't actually doing those things, just something worth knowing for the future

duncanparkinson (changes requested):
Good work on this! The only thing is indentation which needs to be fixed. I know that sounds like just a nit-pick but try to catch these things before they get to review because they can potentially introduce very hard to diagnose bugs

duncanparkinson (approved):

PR #121, opened 2026-04-13
fix/post route is updated with the object the backend expects 

katarzynakaz:
Updated the TraineeBookingFlow component to align with the latest backend API requirements for creating a meeting.

Changes:

Changed the bookingDetailsObj to send volunteer_id, slot_rule_id, time_slot, and agenda.

Removed trainee_name, trainee_email, and manual end_time calculations as the backend now handles these.

Matched time formatting to the backend expectation (YYYY-MM-DDTHH:MM:00Z).

Clearly named variables (combinedDateAndTimeFromUrl, timeSlotForBackend) to avoid confusion later.

Updated Object Format:

JSON
{
    "volunteer_id": 1,
    "slot_rule_id": 1,
    "time_slot": "2026-01-01T10:00:00Z",
    "agenda": "lots of interesting stuff"
}

Note: A follow-up PR will edit BookingForm UI to include the agenda field and remove the previously sent trainee info.


Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -38,16 +38,15 @@ const TraineeBookingFlow = () => {
...
-			trainee_email: bookingFormData.traineeEmail,
-			start_time: startDateTime.toISOString(),
-			end_time: endDateTime.toISOString(),
-			volunteer_name: activeVolunteer.name,
-			volunteer_email: activeVolunteer.email,
+			volunteer_id: activeVolunteer.id,
+			slot_rule_id: 1, //TODO will be updated in the next PR
+			time_slot: timeSlotForBackend,
+			agenda:
+				bookingFormData.agenda || "Tell the volunteer what you need help with.",
```
duncanparkinson: If the trainee doesn't enter an agenda will this not mean we end up saving "Tell the volunteer what you need help with." to the db for the booking?

duncanparkinson (changes requested):
Just a question about the fallback option but otherwise looks good. We're going to need to think about time (and timezone) handling some more but let's keep that for another PR. I think a pairing session on that would be very beneficial

Frontend/src/components/pages/TraineeBookingFlow.jsx
katarzynakaz: Thank you, I forgot to update the str. Done :)

duncanparkinson (approved):

PR #122, opened 2026-04-13
fix-booking form updated with agenda

katarzynakaz:
This PR updates the trainee booking form to provide a meeting agenda while removing redundant trainee fields (which are now handled via auth).

Changes
BookingForm.jsx:
Swapped traineeName and traineeEmail input fields for a greeting using the trainee.name prop.
Added an Agenda textarea with basic validation.
Updated CSS to remove the redundant disabled input.

Note: Even though the email field is no longer required in the Booking Form, I kept the email within the tests to maintain the user object consistent.

### DIFF
--- a/Frontend/cypress/tests/BookingForm.cy.jsx
+++ b/Frontend/cypress/tests/BookingForm.cy.jsx
@@ -8,7 +8,7 @@ describe("BookingForm Component tests", () => {
 
 		cy.get("label.form-label").should(
 			"contain",
-			"Okay Kaska, what would you like to discuss?"
+			"Okay Kaska Test, what would you like to discuss?"
 		);
 	});
 
@@ -25,6 +25,7 @@ describe("BookingForm Component tests", () => {
 
 		cy.get("textarea.form-input").type("Too short");
 		cy.get("button").contains("Book meeting").click();
+		cy.get("@testedSubmit").should("not.have.been.called");
 	});
 
 	it("allows typing in the agenda field", () => {
--- a/Frontend/cypress/tests/BookingForm.cy.jsx
+++ b/Frontend/cypress/tests/BookingForm.cy.jsx
@@ -1,55 +1,42 @@
 import BookingForm from "../../src/components/groups/BookingForm";
 
 describe("BookingForm Component tests", () => {
-	it("shows the tested name and email and verifies email is readOnly", () => {
-		const testedTrainee = { name: "Test Trainee", email: "trainee@test.com" };
+	it("shows the name in label greting", () => {
+		const testedTrainee = { name: "Kaska Test", email: "kaska@test.com" };
 
 		cy.mount(<BookingForm trainee={testedTrainee} whenFormSubmit={() => {}} />);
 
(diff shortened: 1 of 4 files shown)


Frontend/src/components/groups/BookingForm.jsx
```diff
@@ -2,59 +2,45 @@ import React, { useState } from "react";
...
-	//set up local state that will be passed to app
-	const [name, setName] = useState(trainee?.name || "");
-	// Email is now derived directly from props to avoid stale state
-	const currentEmail = trainee?.email || "";
+	const [firstName] = useState(
+		trainee?.name
+			? trainee?.name.split(" ")[0].charAt(0).toUpperCase() +
+					trainee.name.split(" ")[0].slice(1)
+			: ""
+	);
```
duncanparkinson: Unless this is ever going to change it should be a `const` rather than using state.

Assuming that the user's preferred name is the first word split by whitespace is always risky though. [This](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) is a really good resource for all the ways that dealing with names can go wrong. It's safer to stick with either the full name, or just use a less personal form label.

Frontend/src/components/groups/BookingForm.jsx
```diff
@@ -2,59 +2,45 @@ import React, { useState } from "react";
...
-					<label className="form-label">Email</label>
-					<input
-						className="form-input form-input-disabled"
-						type="email"
-						label="Email"
-						value={currentEmail}
-						readOnly
-					/>
-				</div>
 				<br></br>
```
duncanparkinson: Spacing concerns like this are better handled in CSS

Frontend/cypress/tests/BookingForm.cy.jsx
```diff
@@ -1,55 +1,42 @@
...
-
-		cy.get("button").contains("Book meeting").click();
+		cy.get("label.form-label").should(
+			"contain",
+			"Okay Kaska, what would you like to discuss?"
+		);
 	});
 
-	it("cleans name by removing numbers and symbols on submit", () => {
+	it("sends alert when agenda is too short", () => {
```
duncanparkinson, line 15: Should this also be confirming that the form wasn't submitted?

`cy.get("@testedSubmit").should("not.have.been.called");` or similar?

duncanparkinson (changes requested):

Frontend/src/components/groups/BookingForm.jsx
katarzynakaz: Thank you. I updated it to use the full trainee.name.

Frontend/src/components/groups/BookingForm.jsx
katarzynakaz: Good point. I've removed the <br> an mb to the div in CSS.

Frontend/cypress/tests/BookingForm.cy.jsx
katarzynakaz: It absolutely should. Fixed.

duncanparkinson (approved):

PR #123, opened 2026-04-13
Feat/json booking

### LINKED ISSUE #128: Refactor Booking Endpoint with the new JSON structure
Replaced JSON with the new structure of contract Frontend/Backend for Booking.

> {
>     "volunteer_id": 1,
>     "slot_rule_id": 1,
>     "time_slot": "2026-01-01T10:00:00Z",
>     "agenda": "lots of interesting stuff"
> }

Emilianouz:
This PR updates the /api/create-meeting/ endpoint to use the new booking payload based on SlotRule.

The booking flow now aligns with the domain model:

- bookings reference a slot_rule and a specific time_slot
- trainee is derived from authentication instead of being sent from the frontend

**Changes**

- Updated BookingSerializer to accept volunteer_id, slot_rule_id, and time_slot
- Updated CreateMeetingView to use request.user as trainee and compute end_time internally
- Added slot_rule foreign key to Booking model
- Updated tests to match the new payload structure
- Added SlotRule fixture required for serializer validation

### Migration

A new migration adds the slot_rule field to Booking. After pulling the branch, run:

```
python manage.py makemigrations
python manage.py migrate

```
If you encounter local database issues, you may need to reset your local DB:
```
rm db.sqlite3
python manage.py migrate
```
### API Payload update

Replaced JSON with the new structure:

> {
>     "volunteer_id": 1,
>     "slot_rule_id": 1,
>     "time_slot": "2026-01-01T10:00:00Z",
>     "agenda": "lots of interesting stuff"
> }

Close: #128


Backend/core/migrations/0007_booking_slot_rule.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('core', '0006_slotrule_unique_volunteer_start_time_slot_rule'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='booking',
+            name='slot_rule',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='bookings', to='core.slotrule'),
```
duncanparkinson: `ON DELETE CASCADE` means that if we delete a SlotRule any bookings that reference it will also be deleted. Is that what we actually want?

Backend/core/serializers/booking_serializer.py
```diff
@@ -1,54 +1,59 @@
...
+from core.models import User, Booking, SlotRule
 from core.policies.min_booking_window import MinimumBookingWindowPolicy
 from datetime import timedelta
 
 class BookingSerializer(serializers.Serializer):
-    trainee_email = serializers.EmailField()
-    volunteer_email = serializers.EmailField()
-    start_time = serializers.DateTimeField()
+    volunteer_id = serializers.IntegerField()
+    slot_rule_id = serializers.IntegerField()
```
duncanparkinson, line 8: This field is required (by default) here but is allowed to be null in the database. If we need a `slot_rule_id` for a Booking we should also enforce that in the db

duncanparkinson (changes requested):
Good work on this. A couple of data integrity issues to think about though

Backend/core/migrations/0007_booking_slot_rule.py
Emilianouz: I don’t think we want bookings to be deleted if a SlotRule is removed. I'll refactor it.

Backend/core/serializers/booking_serializer.py
Emilianouz: Yes, agreed.
I ran into some migration issues because the database already has existing data (and the DB didn't have that column). I initially added null=True, blank=True just to get the migration to pass, but I agree it should be a required field. I’ll update the migration to include a default or handle the existing nulls so we can enforce this at the DB level.

duncanparkinson: This is the kind of issue that comes up quite often once an app is in production and the existing data needs to be kept. In this case I think we can safely delete the existing data, which can be done in a migration so that it will run for everyone else too (when they run the migration I mean)

Emilianouz: I’ve now updated this so it’s enforced at the DB level as well.

`slot_rule` field is no longer nullable and the model and  I also adjusted the migration.

duncanparkinson: ~~Does the migration handle existing data or will everyone need to delete any existing data for this to work?~~

Never mind, I see it's now in the PR body. Please also post a note about this in the Slack channel though, otherwise people won't know where to look if they have any issues on `main` once this is merged

Backend/core/serializers/booking_serializer.py
```diff
@@ -1,54 +1,59 @@
...
-from core.models import User, Booking
+from core.models import User, Booking, SlotRule
 from core.policies.min_booking_window import MinimumBookingWindowPolicy
 from datetime import timedelta
 
 class BookingSerializer(serializers.Serializer):
-    trainee_email = serializers.EmailField()
-    volunteer_email = serializers.EmailField()
-    start_time = serializers.DateTimeField()
+    volunteer_id = serializers.IntegerField()
```
duncanparkinson: If we're passing in the `slot_rule_id` can we derive the `volunteer_id` from the SlotRule? That would avoid a situation where the two don't agree. Is there a reason not to do that?

duncanparkinson (changes requested):
I've added a question about where we get the `volunteer_id` from.

One other question: do we have a test case for when the `start_time` isn't actually one of the times covered by the SlotRule? We need to make sure that trainees can only book valid slots and that's an example of an invalid one I don't think we currently have a test for?

(If you add any more tests consider carefully whether they should be view tests or service layer tests like we worked on previously. Service layer tests are likely to be easier to write for things like checking whether a time exists in a particular slot_rules' slots, for example)

Backend/core/serializers/booking_serializer.py
Emilianouz: Removed `volunteer_id` from the input and now derive the volunteer from the SlotRule. 
Thank you.

duncanparkinson:
> One other question: do we have a test case for when the `start_time` isn't actually one of the times covered by the SlotRule? We need to make sure that trainees can only book valid slots and that's an example of an invalid one I don't think we currently have a test for?
> 
> (If you add any more tests consider carefully whether they should be view tests or service layer tests like we worked on previously. Service layer tests are likely to be easier to write for things like checking whether a time exists in a particular slot_rules' slots, for example)

Has this question been addressed?

SaraTahir28:
A test has been added to check if start_times arent one of the occurrences in slot_rules.

Emilianouz:
@duncanparkinson 
Included Booking_validation service with slot_rule_covers_time() that returns True if the requested booking time is valid for the given SlotRule.

Behavior:
- one-off rule: exact datetime match
- recurring rule: same time-of-day, on/after start date, on/before repeat_until

Backend/core/services/booking_validation.py
```diff
@@ -0,0 +1,13 @@
...
+def slot_rule_covers_time(slot_rule, time_slot) -> bool:
+
+    if slot_rule.repeat_until is None:
+        return time_slot == slot_rule.start_time
+
+    return (
+        time_slot.time() == slot_rule.start_time.time()
+        and time_slot.date() >= slot_rule.start_time.date()
+        and time_slot.date() <= slot_rule.repeat_until
+    )
```
duncanparkinson: Is this missing a day of the week check? I think this check would allow a slot to be booked for the wrong day as long as it was between the start and end points and at the correct time of day?

duncanparkinson (changes requested):

Backend/core/services/booking_validation.py
Emilianouz: Added a weekday check for recurring rules

duncanparkinson (approved):

PR #124, opened 2026-04-13
Feature/ Integrate Available Slots API in Trainee Booking Flow

katarzynakaz:
This PR implements the fetching and transformation for the frontend of available slot data from the /api/available-slots/ endpoint.
The backend returns 
[
{
"start_time": "2026-01-01T09:00:00Z",
"end_time": "2026-01-01T10:00:00Z",
"volunteer_id": 1,
"slot_rule_id": 1,
"name": "Duncan Parkinson",
"img": "/public/placeholder.png"
}
]
Changes:
Verbose string splitting for date and time to avoid brain meltdown:

String from API: "2026-03-20T09:00:00Z"
Extracting Day Number for Calendar:
let dateOnlyStr = "2026-03-20T09:00:00Z".split("T")[0]; // "2026-03-20"
let dateBitsArr = dateOnlyStr.split("-"); // ["2026", "03", "20"]
let dayNumber = Number(dateBitsArr[2]); // 20

State Management: Added allVolunteersData state to hold the API response.

API call is wrapped in a conditional check on state change (allVolunteersData === null) to only execute once.

Filtering: Data is currently filtered to match activeVolunteer.id (Duncan, ID: 1).

Created convertedAllVDataToFrontendFormat to hold the final processed arrays, ensuring the child components receive the exact types they expect (Numbers for the Calendar, Strings for the TimeSlots).

Testing Instructions
VGET request to /api/available-slots/ is successful.
Confirm the Calendar dots appear based on the dayNumber from the API.
Confirm the TimeSlotGroup buttons display the correct time strings.

Note: Removed local volunteersDetails mock data. The current implementation filters volunteer_id: 1 (Duncan) for the initial API integration. 
Next PR: A separate PR will handle all volunteers.


Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -4,17 +4,79 @@ import SessionDetails from "../groups/SessionDetails";
...
 	const navigate = useNavigate();
 
+	if (allVolunteersData === null) {
+		setAllVolunteersData([]);
+		api
+			.get("/api/available-slots/")
+			.then((res) => setAllVolunteersData(res.data))
+			.catch((err) => {
+				console.log("No api data :():", err);
+			});
```
duncanparkinson: This needs to be a `useEffect`. If we set state like this within the render then it will cause an infinite loop (the null check at line 25 means it will only actually be two loops in most cases, but we still don't want that).

```
useEffect(() => {
    api.get("/api/available-slots/")
        .then((res) => setAllVolunteersData(res.data))
        .catch((err) => console.log(err));
}, []);
```

duncanparkinson (changes requested):
This is definitely going in the right direction. We're going to have some timezone and error handling issues with dates which I've mentioned in another PR I think, but let's deal with that in a pairing session because timezones are one of the most frustrating things to deal with!

Frontend/src/components/pages/TraineeBookingFlow.jsx
katarzynakaz: Thanks for the feedback. I was staying away from useEffect as I vaguely remember that in the initial meeting, Daniel mentioned we should avoid useEffect. I assumed it's in all instances. I see your point about the infinite loop and have pushed the requested fix.

katarzynakaz:
> This is definitely going in the right direction. We're going to have some timezone and error handling issues with dates which I've mentioned in another PR I think, but let's deal with that in a pairing session because timezones are one of the most frustrating things to deal with!

Sounds good. Timezones are very tricky.

Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -1,20 +1,80 @@
...
 import { useAuth } from "../../AuthContext";
 
 const TraineeBookingFlow = () => {
+	const [allVolunteersData, setAllVolunteersData] = useState(null);
+	//TODO next PR show all slots from all volunteers
+	const [activeVolunteer, setActiveVolunteer] = useState({
+		id: 1,
+		name: "Getting volunteer...",
+		img: "/public/placeholder.png",
+	});
```
duncanparkinson: This is potentially confusing I think. We don't want that ID of 1 propagating through the system, and it would be better to have the "Getting volunteer..." text be an explicit loading conditional somewhere instead of initialising a fake user like this

Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -1,20 +1,80 @@
...
 	const { selectedDate, selectedTime, status } = useParams();
 	const { user } = useAuth();
 	const navigate = useNavigate();
 
+	useEffect(() => {
+		api
+			.get("/api/available-slots/")
+			.then((res) => setAllVolunteersData(res.data))
+			.catch((err) => console.log(err));
+	}, []);
```
duncanparkinson, line 35: Similar to above, I think this should probably init to null and show a loading page using a conditional rather than a blank calendar like I think will happen now.

(I know I suggested code that had the empty array initialiser but I've changed my mind, sorry! 😅)

duncanparkinson (changes requested):
A couple more changes needed I think, just to do with how the loading state is handled. Once that is sorted this is good to go!

Frontend/src/components/pages/TraineeBookingFlow.jsx
katarzynakaz: Done, fake user is removed, activeVolunteer initializes null and conditional loading is added.

Frontend/src/components/pages/TraineeBookingFlow.jsx
katarzynakaz: removed as requested

Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -1,20 +1,92 @@
...
+						id: res.data[0].volunteer_id,
+						name: res.data[0].name,
+						img: "/public/placeholder.png",
+					});
+				}
+			})
+			.catch((err) => console.log(err));
+	}, []);
+
+	if (allVolunteersData === null || activeVolunteer === null) {
```
duncanparkinson, line 37: If there isn't any volunteer data this will stay stuck at "loading" forever because `activeVolunteer` will never be set. A check for the `allVolunteersData.length === 0` case would prevent that:

```
if (allVolunteersData === null) {
    return (
        <div className="booking-box">
            <h2>Loading volunteers...</h2>
        </div>
    );
}

if (allVolunteersData.length === 0) {
    return (
        <div className="booking-box">
            <h2>No volunteers are available at this time.</h2>
        </div>
    );
}
```

Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -1,20 +1,92 @@
...
+		});
+
+		if (activeVolunteerSlots.length > 0) {
+			let firstSlot = activeVolunteerSlots[0];
+
+			showActiveVolunteer = {
+				id: activeVolunteer.id,
+				name: firstSlot.name,
+				img: firstSlot.img,
+			};
```
duncanparkinson: By this point in the code we're either displaying a "no volunteers" message or `activeVolunteer` has been initialised in the `useEffect` I think? So do we need `showActiveVolunteer` for anything?

duncanparkinson (changes requested):
A couple more things to look at but otherwise this looks good.

One small thing for PRs in general: use backticks (single or triple depending on context) for any code in the PR body. It preserves the formatting and makes it a lot easier to read. Thanks!

Frontend/src/components/pages/TraineeBookingFlow.jsx
katarzynakaz: Thank you for catching this. My brain was definitely starting to melt by the end of this.

Frontend/src/components/pages/TraineeBookingFlow.jsx
katarzynakaz: Yes, it's redundant, removed as requested.

Frontend/src/components/pages/TraineeBookingFlow.jsx
```diff
@@ -1,20 +1,89 @@
...
+	useEffect(() => {
+		api
+			.get("/api/available-slots/")
+			.then((res) => {
+				setAllVolunteersData(res.data);
+				if (res.data && res.data.length > 0) {
+					setActiveVolunteer({
+						id: res.data[0].volunteer_id,
+						name: res.data[0].name,
+						img: "/public/placeholder.png",
```
duncanparkinson: Can we pull this from the response data? The backend can then send the placeholder for every user, but we won't have to change the frontend in future when there are proper volunteer pictures

duncanparkinson (approved):
One very small change I spotted. Go ahead and merge once that's done ☺️

Frontend/src/components/pages/TraineeBookingFlow.jsx
katarzynakaz: thank you, that's updated

PR #133, opened 2026-04-14
Feature/ connecting frontend create slot to api/slot-rules

katarzynakaz:
This PR connects the frontend to them to /api/slot-rules/. 

Changes
VolunteerDash.jsx matches the agreed contract: volunteer, start_time, repeat_until, and group.
Added a repeat_until date picker to the VolunteerAvailabilityForm for recurring slots.
Updated AddingSlotsBasket to display the end date for recurring rules.
Updated Cypress component integration test (volunteer_flow.cy.js) to add "Repeat until" field.

Testing
npx cypress open
select e2e Integration-VolunteerOnboarding.cy.jsx

Further Changes 
Promise.all: Replaced the for loop with Promise.all combined with .map() so the UI only updates after all slot creation requests have successfully resolved.
The slot data are passed directly into the api.post call instead of creating an object.
The alert and state updates (setHasUserSetAvailability) are set in the .then() block, so the user sees them only when data is received by the backend.
Volunteer is removed because the backend now identifies the volunteer via the active session/token.

### DIFF
--- a/Frontend/src/components/pages/VolunteerDash.jsx
+++ b/Frontend/src/components/pages/VolunteerDash.jsx
@@ -8,9 +8,9 @@ import VolunteerEditSession from "../groups/VolunteerEditSession";
 import VolunteerViewSession from "../groups/VolunteerViewSession";
 
 import VolunteerAvailabilityForm from "../groups/VolunteerAvailabilityForm";
-import { ActionBtn } from "../elements/Button";
 import { useAuth } from "../../AuthContext";
 import duncanImg from "../../assets/duncan.png";
+import api from "../../api/axiosClient";
 
 const VolunteerDash = () => {
 	const { id } = useParams();
@@ -36,12 +36,30 @@ const VolunteerDash = () => {
 
 	const sendVolunteerSlotsToDb = () => {
 		console.log("Ready to send to db", temporaryAddedSlotsStorage);
-		alert("Availability is saved.");
+		for (let i = 0; i < temporaryAddedSlotsStorage.length; i++) {
+			let slotWeAreOn = temporaryAddedSlotsStorage[i];
 
+			let objectToSendToBackend = {
+				volunteer: slotWeAreOn.volunteer,
+				start_time: slotWeAreOn.start_time,
+				repeat_until: slotWeAreOn.repeat_until,
+				group: "all",
+			};
+
+			api
+				.post("/api/slot-rules/", objectToSendToBackend)
+				.then(() => {
+					console.log("Slot zapisany w bazie!");
+					setHasUserSetAvailability(true);
+					setTemporaryAddedSlotsStorage([]);
+				})
+				.catch((error) => console.log("Error:", error));
+		}
+
+		alert("Availability is saved.");
 		setHasUserSetAvailability(true);
 	};
 
-	// grab all sessions for active user
 	const activeVolunteerSessions = allBookedSessionsForAllUsers.filter(
 		(session) => session.volunteerId === activeVolunteer.id
 	);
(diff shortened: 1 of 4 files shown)


Frontend/src/components/pages/VolunteerDash.jsx
```diff
@@ -36,12 +36,30 @@ const VolunteerDash = () => {
...
+				.post("/api/slot-rules/", objectToSendToBackend)
+				.then(() => {
+					console.log("Slot zapisany w bazie!");
+					setHasUserSetAvailability(true);
+					setTemporaryAddedSlotsStorage([]);
+				})
+				.catch((error) => console.log("Error:", error));
+		}
 
+		alert("Availability is saved.");
```
duncanparkinson: This alert will show no matter what: it needs to be in the `then` block instead

Frontend/src/components/pages/VolunteerDash.jsx
```diff
@@ -36,12 +36,30 @@ const VolunteerDash = () => {
...
 		setHasUserSetAvailability(true);
```
duncanparkinson: This is already called in the `then` block so we don't need it here

Frontend/src/components/pages/VolunteerDash.jsx
```diff
@@ -36,12 +36,30 @@ const VolunteerDash = () => {
 
 	const sendVolunteerSlotsToDb = () => {
 		console.log("Ready to send to db", temporaryAddedSlotsStorage);
-		alert("Availability is saved.");
+		for (let i = 0; i < temporaryAddedSlotsStorage.length; i++) {
+			let slotWeAreOn = temporaryAddedSlotsStorage[i];
+
+			let objectToSendToBackend = {
+				volunteer: slotWeAreOn.volunteer,
```
duncanparkinson: The backend is going to take this from the session (logged in user) rather than the request I think?

Frontend/src/components/pages/VolunteerDash.jsx
```diff
@@ -36,12 +36,30 @@ const VolunteerDash = () => {
...
+
+			api
+				.post("/api/slot-rules/", objectToSendToBackend)
+				.then(() => {
+					console.log("Slot zapisany w bazie!");
+					setHasUserSetAvailability(true);
+					setTemporaryAddedSlotsStorage([]);
+				})
+				.catch((error) => console.log("Error:", error));
+		}
```
duncanparkinson: There are some async issues here (one of the things that can be quite confusing/different about frontend work). It's difficult to get something like a for-loop to play nicely with anything you want to have happen after the entire loop is complete if the items in the loop are all async. The solution is to use `Promise.all` ([here's a really good StackOverflow answer explaining this](https://stackoverflow.com/questions/38180080/when-to-use-promise-all))

There's a potential gotcha with this though, which is that `forEach` will return `undefined` so any loop actually needs to be a `map` so that the promises are correctly returned to `Promise.all`.

It's also possible to inline the object being sent to the backend (which is easier to read in most cases). So with all of that combined the final code would look something like:
```
const sendVolunteerSlotsToDb = () => {
    Promise.all(
        temporaryAddedSlotsStorage.map((slot) =>
            api.post("/api/slot-rules/", { start_time: slot.start_time, repeat_until: slot.repeat_until, group: "all" })
        )
    )
    .then(() => {
        setHasUserSetAvailability(true);
        setTemporaryAddedSlotsStorage([]);
        alert("Availability is saved.");
    })
    .catch((error) => {
        console.error("Error saving slots:", error);
        alert("Failed to save availability. Please try again.");
    });
};
```

duncanparkinson (changes requested):

Frontend/src/components/pages/VolunteerDash.jsx
katarzynakaz: repeated code deleted

Frontend/src/components/pages/VolunteerDash.jsx
katarzynakaz: thank you, it's updated as requested

Frontend/src/components/pages/VolunteerDash.jsx
katarzynakaz: moved to then block

duncanparkinson (approved):

PR #135, opened 2026-04-14
Add deleted_at field to SlotRule and update unique constraint logic

### LINKED ISSUE #134: SlotRule deleted_at timestamp implementation.
Implement soft delete with: `deleted_at`  to encode two pieces of information  in a single field: null values mean it is still active.

- SlotRule model, 
- update endpoint to "delete"
- update Constraints, 
- available_slot filtering

Emilianouz:
**User story:** As a volunteer, I want to delete an offered time slot, so that I can remove availability I can no longer fulfil

This PR introduces a soft delete pattern for the SlotRule model: records are marked as inactive using a deleted_at timestamp. 

### Changes

- Added deleted_at field to SlotRule to support soft delete
- Updated unique constraint to apply only to active (deleted_at IS NULL) slot rules
- Created migration for deleted_at field and updated constraint
- Updated SlotRuleSerializer validation to ignore soft-deleted records
- Implemented soft delete endpoint:

     - DELETE /slot-rules/{id}/
    - Sets deleted_at instead of deleting the record
    - Restricts deletion to the owner (returns 403 otherwise)

- Updated AvailableSlotsView to exclude soft-deleted slot rules

- Added tests for:

    - Soft delete behavior on slot rule deletion
    - Ownership restriction on delete (403 response)
    - Creating slot rules when previous ones are soft-deleted
    - Ensuring soft-deleted slot rules are not returned in available slots

**API - Endpoint**
Delete SlotRule (Soft Delete)

> DELETE `/slot-rules/{id}/`

**Testing:** 
Updated existing tests to account for soft delete behavior

Instructions:
`
python manage.py migrate
`

Check list:

- [x] A backend endpoint exists to delete a slot rule
- [x] The deletion is a soft delete, using a column like deleted_at or similar
- [x] A volunteer can only delete slot rules they own — deleting another volunteer's slot returns 403
- [x] Deleted slots no longer appear in the trainee availability view
- [ ] The volunteer dashboard provides UI to delete a published slot rule
- [x] Existing bookings are left intact

UI out of scope for this PR.

close #134


Backend/core/tests/test_slot_rule_delete_view.py
```diff
@@ -0,0 +1,148 @@
...
+        slot_rule = SlotRule.objects.create(
+            volunteer=volunteer,
+            start_time="2026-04-12T10:00:00Z",
+            repeat_until="2026-04-20",
+            group="sdc",
+        )
+
+        response = api_client.delete(self.get_url(slot_rule.id))
+
+        assert response.status_code in [401, 403]
```
duncanparkinson: If the user is unauthenticated is there any way for 403 to happen here? If not we should just assert 401

Backend/core/views.py
```diff
@@ -142,3 +143,16 @@ class SlotRuleCreateView(generics.CreateAPIView):
     # Always assign the slot rule to the logged-in user
     def perform_create(self, serializer):
         serializer.save(volunteer=self.request.user)
+
+class SlotRuleDeleteView(generics.DestroyAPIView):
+    queryset = SlotRule.objects.filter(deleted_at__isnull=True)
+    serializer_class = SlotRuleSerializer
```
duncanparkinson, line 149: Are we actually using this? If not this is dead code and can be removed

duncanparkinson (approved):
Looks good apart from a couple of very small things ☺️ 

One other question: the checklist has UI on it but then you also say that's out of scope? If it's intentionally out of scope then this is ready to merge once the small changes are made

Backend/core/views.py
Emilianouz: It's used. It's the API endpoint to "delete" (soft delete) SlotRules.
`http://localhost:8000/api/slot-rules-delete/1/`
It works correctly.

duncanparkinson: That's the view generally, I was asking specifically about line 149 where it assigns `serializer_class`. It doesn't look like that assignment is ever used, unless I'm missing something?

[edit to add]: I now see that I selected too many lines when I added the first comment which is the source of the confusion! It should only have highlighted line 149

Backend/core/tests/test_slot_rule_delete_view.py
Emilianouz: Good point.  I agree we should assert 401 explicitly.

duncanparkinson: It looks like DRF has changed how this works and does throw 403 in this situation. Ideally we'd assert only one status code because we should know how we expect it to fail. In this case it looks like it should actually be just 403. It's not a massive issue so let's leave it for now

Backend/core/views.py
Emilianouz: It's required by DRF generic views, DestroyAPIView does not use it for response serialization but it is still part of the view contract.

[`serializer_class` - The serializer class that should be used for validating and deserializing input, and for serializing output.](https://www.django-rest-framework.org/api-guide/generic-views/#attributes)

Backend/core/urls.py
```diff
@@ -13,6 +13,6 @@
     path("users/<int:pk>/", UserDetailView.as_view(), name="user-detail"),
     path("available-slots/", AvailableSlotsView.as_view(), name="available-slots"),
     path("profile/", CurrentProfileView.as_view(), name="current-profile"),
-    path("slot-rules/", SlotRuleCreateView.as_view(), name="slot-rule-create"),
-    path("slot-rules/<int:pk>/", SlotRuleDeleteView.as_view(), name="slot-rule-delete"),
+    path("slot-rules-create/", SlotRuleCreateView.as_view(), name="slot-rule-create"),
+    path("slot-rules-delete/<int:pk>/", SlotRuleDeleteView.as_view(), name="slot-rule-delete"),
```
duncanparkinson, line 17: These were correct as they were. The delete is handled through the HTTP method (`POST`, `GET`, `DELETE`, etc) so putting it in the URL itself is non-standard and not correct for REST. It will also break any frontend code relying on the URLs the way they were. Best to just revert this commit

Backend/core/views.py
duncanparkinson: That still relies on the serializer actually being invoked I believe. If you remove the line does it break or does everything still work?

Backend/core/urls.py
Emilianouz: That's totally true! I don't know what I was thinking. Thanks for catching that. I've updated it to follow REST standards.

Backend/core/tests/test_slot_rule_delete_view.py
Emilianouz: Thank you. It's just aligning with [the acceptance criteria](https://github.com/SaraTahir28/Pair-Scheduling-APP/issues/139)

Issue #139, opened 2026-04-14
Volunteers can delete their offered time slots


duncanparkinson:
**User story:** As a volunteer, I want to delete an offered time slot, so that I can remove availability I can no longer fulfil

**Acceptance criteria:**
- [x] A backend endpoint exists to delete a slot rule
- [x] The deletion is a soft delete, using a column like `deleted_at` or similar
- [x] A volunteer can only delete slot rules they own — deleting another volunteer's slot returns 403
- [x] Deleted slots no longer appear in the trainee availability view
- [x] The volunteer dashboard provides UI to delete a published slot rule
- [x] Existing bookings are left intact

**Out of scope:**
- Notifying affected trainees

duncanparkinson:
#150 is very similar to this but different in implementation

Issue #138, opened 2026-04-14
Volunteers can edit their offered time slots


duncanparkinson:
**User story:** As a volunteer, I want to edit an offered time slot, so that I can correct mistakes or update my availability

**Acceptance criteria:**
- [ ] A backend endpoint exists to update a slot rule
- [ ] Updates should be handled as a "delete" (see #139) then creating a new replacement slot
- [ ] A volunteer can only edit slot rules they own — editing another volunteer's slot returns 403
- [ ] The volunteer dashboard provides UI to edit a published slot rule
- [ ] Changes are reflected immediately in the trainee availability view

**Out of scope:**
- Notifying trainees of the change
- Cancelling bookings affected by the change
- Displaying a list of slot rules is a separate issue, this issue can just use direct URL access to the edit page if it is done before the list view

PR #147, opened 2026-04-15
Feature/ integration of volunteer availability with API

katarzynakaz:
This PR implements the complete flow for managing volunteer availability across onboarding, dashboard, and availability manager. The PR is large due to data flow between the onboarding form, volunteer dashboard, and the new availability manager components.

Changes
- Connected VolunteerDash to fetch volunteer availability from the `/api/slot-rules/` endpoint.
- Added VolunteerAvailabilityManager.
- POST identifies existing records and only POSTs new slots.
- AvailabilityManager is accessible from SessionDetails. 
- Existing slot rules can be removed with `DELETE /api/slot-rules/:id/`

Availability management behaviour
- Existing slot rules are fetched from the API and displayed
- New slot rules are created with `POST /api/slot-rules/`
- Current edit flow supports adding and removing slot rules rather than in-place updating of an existing rule

Testing:
npx cypress open 
component integration test - tests/SlotManagement.cy.jsx, e2e volunteer onboarding.

Notes
- "Upcoming sessions" is kept as a placeholder section for presentation / future development - To be discussed with the team.


Frontend/src/components/groups/AddingSlotsBasket.jsx
```diff
@@ -2,21 +2,27 @@ import React from "react";
...
 	return (
 		<div className="basket-container">
-			<h3 className="basket-title">Entries to save:</h3>
+			<h3 className="basket-title">{title}</h3>
 
 			<div className="basket-list">
 				{addedSlots.map((entry, index) => (
 					<div className="basket-row" key={index}>
-						<div className="flex items-center gap-4">
+						<div className="basket-entires">
```
duncanparkinson: just a typo: should be "basket-entries"?

duncanparkinson (changes requested):
Functionality looks good for most of this. At the moment recurring slots are showing up correctly when saved but then incorrectly once I reload though (see below). It looks like on a full page load the frontend is displaying all of the slots when it should be displaying the slot rules instead?

I also think the volunteer dash should probably just go straight to the "manage my availability" page? The "upcoming bookings" section is currently hardcoded and it isn't on the must haves (at least I don't think it is, worth checking the [spec](https://curriculum.codeyourfuture.io/launch/projects/pair-scheduling/) for this). It would be nice to have it, but we need to prioritise getting all the must haves in place first.

Image

duncanparkinson:
> Functionality looks good for most of this. At the moment recurring slots are showing up correctly when saved but then incorrectly once I reload though (see below). It looks like on a full page load the frontend is displaying all of the slots when it should be displaying the slot rules instead?
> 
> I also think the volunteer dash should probably just go straight to the "manage my availability" page? The "upcoming bookings" section is currently hardcoded and it isn't on the must haves (at least I don't think it is, worth checking the [spec](https://curriculum.codeyourfuture.io/launch/projects/pair-scheduling/) for this). It would be nice to have it, but we need to prioritise getting all the must haves in place first.
> 
> Image

Though considering this further, we need to decide how #139 and #150 interact with each other and how the interface should look. The current interface would work for removing individual slots I think (#150) but not for #139. #138 also requires an edit interface for the slot rules itself, which on the backend will be handled as a deactivation and then the creation of a new slot rule.

Frontend/src/components/groups/AddingSlotsBasket.jsx
katarzynakaz: fixed thank you

katarzynakaz:
> > Functionality looks good for most of this. At the moment recurring slots are showing up correctly when saved but then incorrectly once I reload though (see below). It looks like on a full page load the frontend is displaying all of the slots when it should be displaying the slot rules instead?
> > I also think the volunteer dash should probably just go straight to the "manage my availability" page? The "upcoming bookings" section is currently hardcoded and it isn't on the must haves (at least I don't think it is, worth checking the [spec](https://curriculum.codeyourfuture.io/launch/projects/pair-scheduling/) for this). It would be nice to have it, but we need to prioritise getting all the must haves in place first.
> > Image
> 
> Though considering this further, we need to decide how #139 and #150 interact with each other and how the interface should look. The current interface would work for removing individual slots I think (#150) but not for #139. #138 also requires an edit interface for the slot rules itself, which on the backend will be handled as a deactivation and then the creation of a new slot rule.

I’ve updated the PR with the following fixes:
Changed fetch from available-slots to slot-rules and renamed all variables for clarity, fixed the reload issue you mentioned, fixed the CSS typo and updated Cypress to match /api/slot-rules/

Regarding the Dashboard: I suggest keeping the "Upcoming sessions" section as a placeholder for our presentation. It is to cover points for future development and to show the actual bookings in the future UI.

Frontend/src/components/groups/VolunteerAvailabilityManager.jsx
```diff
@@ -0,0 +1,167 @@
...
+		const slotRuleToRemoveFromApiOrUi = slotRulesInBasketDraft[indexToRemove];
+		if (slotRuleToRemoveFromApiOrUi.id) {
+			api
+				.delete(`/api/slot-rules/${slotRuleToRemoveFromApiOrUi.id}/`)
+				.then(() => {
+					const updatedSlotRuleBasketNotYetSent = slotRulesInBasketDraft.filter(
+						(_, i) => i !== indexToRemove
+					);
+					setSlotRulesInBasketDraft(updatedSlotRuleBasketNotYetSent);
+					setMyOldSlotRulesFromApi(updatedSlotRuleBasketNotYetSent);
```
duncanparkinson: I'm a bit confused by this naming. This seems to be updating a variable that is explicitly named for the old data? Is that right?

Frontend/src/components/pages/VolunteerDash.jsx
```diff
@@ -33,6 +34,23 @@ const VolunteerDash = () => {
 	const volunteerSubmitedFormWithSlots = (newSlotObj) => {
 		setTemporaryAddedSlotsStorage([...temporaryAddedSlotsStorage, newSlotObj]);
 	};
+	const [showManager, setShowManager] = useState(false);
+
+	useEffect(() => {
+		if (activeVolunteer?.id) {
+			api
+				.get("/api/slot-rules/")
```
duncanparkinson, line 42: Is this data being fetched twice? Here and in VolunteerAvailabilityManager?

Frontend/src/components/groups/VolunteerAvailabilityManager.jsx
```diff
@@ -0,0 +1,167 @@
...
+								title="Current availability"
+							/>
+						)}
+						{myOldSlotRulesFromApi.length === 0 && (
+							<div className="basket-container">
+								<h3 className="basket-title">Current availability</h3>
+								<div className="basket-list">
+									<p>
+										No availability sessions set. You can add some by clicking
+										`Edit`.
```
duncanparkinson: Can these be single quotes instead of backticks please?

duncanparkinson (changes requested):
What's the intended workflow for the list vs the 'add new' part? At the moment clicking the cross immediately deletes but adding something new to the list requires clicking the 'save all' button. That inconsistency is a little confusing to me.

Frontend/src/components/groups/VolunteerAvailabilityManager.jsx
katarzynakaz: done

Frontend/src/components/pages/VolunteerDash.jsx
katarzynakaz: Yes, it is currently fetched twice. VolunteerDash fetches it for hasUserSetAvailability conditional rendering of the Onboarding form or he main view of the dash with upcoming booked sessions. The VolunteerAvailabilityManager fetches it to keep its own state for editing. I kept it independent for the MVP and added to the refactoring task https://github.com/users/SaraTahir28/projects/13?pane=issue&itemId=171360687&issue=SaraTahir28%7CPair-Scheduling-APP%7C76

Frontend/src/components/groups/VolunteerAvailabilityManager.jsx
katarzynakaz: Thank you, this was wrong and was overwriting the API state, it updates the draft now

katarzynakaz:
> What's the intended workflow for the list vs the 'add new' part? At the moment clicking the cross immediately deletes but adding something new to the list requires clicking the 'save all' button. That inconsistency is a little confusing to me.

That was my oversight, thank you for flagging it.
I've aligned the behaviour so both adding and removing slots are treated as pending changes and applied on "Save all".

Frontend/eslint.config.js
```diff
@@ -1,57 +1,39 @@
...
-        context: 'readonly',
-        it: 'readonly',
-        before: 'readonly',
-        beforeEach: 'readonly',
-        after: 'readonly',
-        afterEach: 'readonly',
-        expect: 'readonly',
-        assert: 'readonly',
-      },
-    },
```
duncanparkinson, line 26: We want this section, it's what stops the errors in Cypress files

Frontend/cypress/e2e/Integration-VolunteerOnboarding.cy.jsx
```diff
@@ -1,5 +1,16 @@
+/* eslint-disable no-undef */
```
duncanparkinson: We shouldn't need this if the options I've highlighted in `eslint.config.js` are restored

Frontend/cypress/tests/SlotManagement.cy.jsx
```diff
@@ -0,0 +1,61 @@
...
+		cy.contains("Add to list").click();
+
+		cy.contains("Save all").click();
+
+		cy.wait("@saveAction");
+		cy.contains("My availability").should("be.visible");
+	});
+});
+
+export default VolunteerAvailabilityManager;
```
duncanparkinson: I don't think we need this export?

duncanparkinson (changes requested):
Just a couple of small changes needed I think ☺️

Frontend/cypress/tests/SlotManagement.cy.jsx
katarzynakaz: removed the export

Frontend/cypress/e2e/Integration-VolunteerOnboarding.cy.jsx
katarzynakaz: removed /* eslint-disable no-undef */

duncanparkinson (approved):

PR #160, opened 2026-04-15
Feature/authentication by default

SaraTahir28:
This PR applies a global IsAuthenticated requirement to all views and updates the test suite accordingly to get the app production ready security wise.

**Changes included:**
Views

Added IsAuthenticated permission to all views to enforce authenticated access.
Ensured unauthenticated requests consistently return 403 Forbidden.

**Tests**

1. Updated LogoutView tests to expect 403 for unauthenticated requests.
2. Adjusted UserView tests: authenticated requests return 200, unauthenticated requests return 403.
3. Modified UserDetailView tests to explicitly authenticate before each request.
4. Fixed UserListCreateView tests to authenticate before requests and updated assertions to account for the authenticated user being included in results.

### DIFF
--- a/Backend/auth_app/views/loginlogout.py
+++ b/Backend/auth_app/views/loginlogout.py
@@ -1,30 +1,35 @@
-from django.http import JsonResponse
-from django.contrib.auth.decorators import login_required
+from django.middleware.csrf import get_token
 from django.contrib.auth import logout
-from functools import wraps
+from django.contrib.auth import logout
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from rest_framework import status
+from rest_framework.permissions import IsAuthenticated
 
 
-def login_required_json(view_func):
-    @wraps(view_func)
-    def wrapper(request, *args, **kwargs):
-        if not request.user.is_authenticated:
-            return JsonResponse({"error": "Unauthorized"}, status=401)
-        return view_func(request, *args, **kwargs)
-    return wrapper
+class UserView(APIView):
+    permission_classes = [IsAuthenticated]
 
-@login_required_json
-def user_view(request):
-   
-    return JsonResponse({
-        "id": request.user.id,
-        "email": request.user.email or "",
-        "name": request.user.get_full_name() or request.user.username,
-        "role": request.user.role,
-        "status": request.user.status,
-    })
+    def get(self, request):
+        return Response(
+            {
+                "id": request.user.id,
+                "email": request.user.email or "",
+                "name": request.user.get_full_name() or request.user.username,
+                "role": getattr(request.user, "role", None),
+                "status": getattr(request.user, "status", None),
+            }
+        )
 
-def logout_view(request):
-    if request.method == "POST":
+
+class LogoutView(APIView):
+    permission_classes = [IsAuthenticated]
+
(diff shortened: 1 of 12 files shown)


Backend/core/tests/test_user_list_create_view.py
```diff
@@ -5,8 +5,17 @@
 
 @pytest.mark.django_db
 class TestUserListCreateView:
-
     def test_list_users_returns_ordered_list(self, client):
```
duncanparkinson: This test doesn't assert the order. If we don't care about the ordering that's fine, but the name should change to match please.

Backend/auth_app/views/loginlogout.py
```diff
@@ -1,30 +1,34 @@
...
         logout(request)  # clears session
-        return JsonResponse({"message": "Logged out"})
-    return JsonResponse({"error": "Method not allowed"}, status=405)
\ No newline at end of file
+        return Response({"message": "Logged out"}, status=status.HTTP_200_OK)
+
+    def get(self, request):
+        return Response(
+            {"error": "Method not allowed"}, status=status.HTTP_405_METHOD_NOT_ALLOWED
+        )
```
duncanparkinson, line 34: We don't need this. DRF will automatically reject `GET` requests if there is no method

duncanparkinson (changes requested):
A couple of small changes I've noted.

The main thing is there aren't any tests for the primary claim of the PR that it adds auth everywhere. `permission_classes = [IsAuthenticated]` needs to be removed from all places using it and then some test assertions that auth still works on those endpoints. For now just do one as proof that everything is hooked up, adding more thorough testing will need to be a future concern.

Backend/core/tests/test_user_list_create_view.py
SaraTahir28: Thanks done

Backend/auth_app/views/loginlogout.py
SaraTahir28: Removed it

SaraTahir28:
I have  implemented the requested changes, but the login flow is failing after rebasing with main, I am consistently seeing Third‑Party Login Failure and sometimes 403 Forbidden errors when attempting Google OAuth login.
From my investigation, I suspect the issue is related to multiple .env files now present in the project. It looks like there may be a mismatch between which URLs are being loaded (local vs production), and this could be causing the callback validation to fail. I’ve tried adjusting the values in different places, but the problem persists.

duncanparkinson:
You should have:

`.env`: used by Docker Compose
`Backend/.env`: Django secrets including OAUTH
`Backend/.env.development`: sets frontend URL for CORS/CSRF
`Backend/.env.production`: sets frontend URL for CORS/CSRF
`Frontend/.env.development`: sets VITE_API_URL for API calls
`Frontend/.env.production`: sets VITE_API_URL for API calls

The dev and production files should not conflict in any way when using Docker. Are you using `make dev` to run the servers?

duncanparkinson:
The most likely cause is you have entries in .env files from how this worked before that are now conflicting. In particular make sure the root `.env` has only the Postgres stuff and the Google service account

duncanparkinson (approved):

PR #174, opened 2026-04-17
Fix trainee booking slot display and tests for UK local time

Emilianouz:
**Description**
This PR updates the trainee booking flow to display available volunteer slots **in UK local time** instead of using raw UTC values from the API.

The backend returns slot timestamps **in UTC**, for example:

```
{
  "start_time": "2026-01-01T09:00:00Z"
}
```

Previously, the frontend was extracting date and time by splitting the ISO string, which meant no timezone conversion was happening. This caused incorrect slot display when UK daylight saving time applied.

This change converts slot timestamps using Date and formats them explicitly in the Europe/London timezone, so the UI consistently shows UK local time and handles GMT/BST correctly.

**Changes:**

- Replaced string-splitting logic for slot dates/times with proper Date parsing.
- Converted available slot dates using UK timezone-aware formatting.
- Converted displayed slot times using Europe/London.
- Updated selected date matching logic so it compares against UK-local dates instead of UTC date fragments.
- Fixed Cypress tests that were assuming displayed times always matched raw UTC values.
- Updated test expectations to derive displayed time/date in the same timezone as the app.
- Reviewed booking submission behavior so UTC payload expectations stay aligned with route/display timezone behavior.

**Example:**

> API returns 2026-07-01T**09:00**:00Z
> Old UI showed **09:00**
> New UI correctly shows **10:00** in the UK

Result

- [x] Slot display is now correct for UK users
- [x] BST/GMT transitions are handled automatically
- [x] Tests now reflect actual frontend behavior instead of assuming UTC display


duncanparkinson (commented):
The rest of the date and time issues in the codebase are handled without a library dependency. Do we need a new library for this issue?

If the timezone handling expanded to include manually chosen timezones then I'm sure we'd need it, but given how much has already been solved (and is tested) with the existing approach can we double check that this lib is really required at this point please?

Edit to add: I see that it’s pinning the timezone to UK local time. The rest of the code will support the user’s browser timezone, whereas this will force UK time. What’s the rationale for that decision?

PR #179, opened 2026-04-21
Add CSRF and session cookie domain to support production

SaraTahir28:
Fix csrf and session cookie domain to support production.


duncanparkinson (approved):
LGTM — just be careful of PR title length and spelling/grammar ☺️

PR #185, opened 2026-05-27
Feature/Authorization

### LINKED ISSUE #183: Feature- Authorization/ Change to admin must be restricted
**Background**
Currently, the patch request to make someone an admin can be made by any user, which is a risk to the system

**Scenario: admin creation restriction**
GIVEN a user is set up as a user (trainee/volunteer)
WHEN a patch request from an admin to change admin rights is submitted
THEN the user is made an admin

**Scenario: admin creation restriction - error message**
GIVEN a user is set up as a user (trainee/volunteer)
WHEN a patch request to change to admin is done without the necessary authorisation
THEN the error message 403 Forbidden should show

SaraTahir28:
This PR close #183  and does backend authorization by ensuring that only admin users can modify user roles, and that non‑admin users cannot assign or remove the admin role. The update closes a security gap where a malicious user could previously bypass the frontend and escalate privileges via direct API calls.

1.**Serializer Update**
 UserSerializer updated to enforce strict role‑change rules at the backend level.

The serializer now
**Prevents non‑admin users from:**

1. Assigning the admin role to themselves or others
2. Demoting an existing admin

**Allows admin users to**

1. Promote any user to admin
2. Demote any admin to a non‑admin role

2.**Test Suite**

1. Trainee/volunteer cannot promote themselves to admin
2. Trainee/volunteer cannot promote any other user to admin
3. Trainee/volunteer cannot demote an admin
4. Admin can promote a user to admin
5. Admin can demote another admin

All tests pass successfully

3. **Manual Security Verification (Browser DevTools)**
To validate the backend protection beyond automated tests, we manually attempted to bypass the frontend using direct PATCH requests executed via fetch() in the browser DevTools console.

**What we verified:**

1. The browser sends valid session + CSRF cookies
2. Unauthorized users receive 403 Forbidden when attempting to escalate privileges
3. Admin users can successfully promote/demote roles
4. The backend blocks privilege escalation even when the UI is bypassed

This confirms that the API is now protected against direct privilege‑escalation attempts.

**Expected Behavior** 

1. Admin RBAC is enforced 
2. All automated tests pass
3. Manual DevTools attacks confirm the system is secure
4. The API is now protected against privilege‑escalation attempts

### DIFF
--- a/Backend/core/tests/test_authorization.py
+++ b/Backend/core/tests/test_authorization.py
@@ -71,3 +71,31 @@ def test_admin_can_promote_user_to_admin(api_client):
 
     assert response.status_code == 200
     assert response.data["role"] == "admin"
+
+
+@pytest.mark.django_db
+def test_admin_can_demote_an_admin(api_client):
+    admin = User.objects.create_user(
+        username="main_admin",
+        email="main_admin@example.com",
+        password="securepass123",
+        role="admin",
+        status="active",
+    )
+
+    other_admin = User.objects.create_user(
+        username="other_admin",
+        email="other_admin@example.com",
+        password="securepass123",
+        role="admin",
+        status="active",
+    )
+
+    # Authenticate as the main admin
+    api_client.force_authenticate(user=admin)
+    # Admin demotes the other admin to volunteer
+    url = reverse("user-detail", args=[other_admin.id])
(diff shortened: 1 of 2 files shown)


katarzynakaz (approved):
This looks awesome, Sara! Kudos on the detailed description and being so thorough with the testing.

PR #193, opened 2026-06-04
Fix/slots basket bug #187

katarzynakaz:
## Description
This PR addresses issue 187 - https://github.com/SaraTahir28/Pair-Scheduling-APP/issues/187

### Fixes:
1. **Duplicate Entries:** Added validation to prevent volunteers from adding identical time slots.
2. **Overlapping Times:** Implemented a rule of a minimum 1-hour interval between slots to prevent overlapping sessions.
3. **Saving Availability Bug:** Fixed a bug where deleting a slot from the basket and saving would fail due to stale state issues.

### Changes:
- In VolunteerAvailabilityForm added checkForDuplicateSlot and checkForOverlapSlot.
- Added tests to test both.
- In VolunteerAvailabilityManager updated removeSlotFromBasket to fix the stale state bug.

### To test:
- Log in as a Volunteer.
- Add overlapping and duplicate slots

### Note on this branch
This branch is based on an unmerged branch that includes styling fixes.


SaraTahir28 (commented):
Good start on removing the exact duplicates 🙂
I tested the behaviour again and the issue is still partly there.

The new check only blocks identical recurring slots (same start_time + same repeat_until), but the original bug also involves overlapping recurring availability, not just exact matches.

For example:
recurring 14 June → 5 July at 09:00
adding another recurring slot for the same date range at 09:30 is blocked

but individual slots on 21/06 or 5/07 at 09:00 /9:30 can still be added
we still need to catch overlapping recurring rules in addition to identical ones.

Image

Frontend/cypress/tests/VolunteerAvailabilityForm.cy.jsx
```diff
@@ -25,4 +25,54 @@ describe("VolunteerAvailabilityForm", () => {
...
   });
+
+  describe("starting with a slot in basket", () => {
+    beforeEach(() => {
+      const onSubmit = cy.stub().as("onSubmit");
+
+      const alertStub = cy.stub().as("alert");
+      cy.on("window:alert", alertStub);
+
+      const existingSlot = {
```
duncanparkinson: Can we shape it like the real response and add a recurring case please?

```js
const existingSlot = {
  id: 101,
  start_time: new Date(2026, 4, 20, 10, 0).toISOString(),
  repeat_until: "2026-06-08",
  group: "all",
};
```

Frontend/src/components/groups/VolunteerAvailabilityForm.jsx
```diff
@@ -42,6 +42,81 @@ const VolunteerAvailabilityForm = ({
       startTime
     ).toISOString();
 
+    const checkForDuplicateSlot =
+      addedSlots &&
+      addedSlots.some((oldSlot) => {
+        const isSameStartDateAndTime = oldSlot.start_time === timeWithDate;
+
+        if (isRecurring) {
+          const isOldRecurring = oldSlot.regular === true;
```
duncanparkinson: `SlotRuleSerializer` returns `id`, `volunteer`, `start_time`, `repeat_until` and `group`—there's no `regular` field, and `startEditing` seeds the basket straight from that response, so this is `undefined` for every saved slot.

```suggestion
          const isOldRecurring = oldSlot.repeat_until != null;
```

Frontend/src/components/groups/VolunteerAvailabilityForm.jsx
```diff
@@ -42,6 +42,81 @@ const VolunteerAvailabilityForm = ({
...
+
+        const oldSlotDate = new Date(oldSlot.start_time);
+        const newSlotDate = new Date(timeWithDate);
+        const isSameWeekday = oldSlotDate.getDay() === newSlotDate.getDay();
+        const checkMinutesDifference = Math.abs(
+          oldSlotDate.getHours() * 60 +
+            oldSlotDate.getMinutes() -
+            (newSlotDate.getHours() * 60 + newSlotDate.getMinutes())
+        );
+        if (oldSlot.regular && !isRecurring) {
```
duncanparkinson: Same as above

```js
const isOldRecurring = oldSlot.repeat_until != null;
```

Frontend/src/components/groups/VolunteerAvailabilityForm.jsx
```diff
@@ -42,6 +42,81 @@ const VolunteerAvailabilityForm = ({
...
+          return (
+            isSameWeekday &&
+            checkMinutesDifference < 60 &&
+            newSlotDate >= oldSlotDate &&
+            newSlotDate <= new Date(oldSlot.repeat_until)
```
duncanparkinson: `repeat_until` is a date-only string, so `new Date("2026-07-06")` is UTC midnight and any start time on the final recurrence day compares greater and slips through. We already have `parseLocalDate` in `utilities/dateTime` which should cover this ☺️

duncanparkinson (changes requested):
Looks good, just a couple of suggestions to fix :)

PR #194, opened 2026-06-04
Fix/add missing polling

katarzynakaz:
## Description
This PR adds missing long-polling to keep slot availability updated.

### Changes:
- **VolunteerAvailabilityManager**: Added polling to fetch /api/slot-rules/.
- **TraineeBookingFlow**: Added polling to /api/available-slots/.

### Note:
Please note this branch is built on top of a branch with fixes.


duncanparkinson (approved):
This looks good! One thing to change: can we disable the polling while the user is editing? We don't want to change the data out from under them.

We probably don't want to use `setInterval` long term for periodic fetching, but let's defer that decision to later!

PR #196, opened 2026-06-05
Fix visibility of buttons - manage volunteer availability and save all in the slot basket

### LINKED ISSUE #190: Fix/hide manage availability button on onboarding
During first-time login, the availability form opens automatically, but the “Manage Availability” button is still visible even though the user is already inside that flow.
Slot basket bug - save all button disappears.

katarzynakaz:
This PR addresses two issues:
1. Manage Availability button visibility
The button now remains visible but becomes "active" (disabled state, distinct styling) when the user is inside the availability manager.

2. Slot basket Save all button disappearing bug
Fixed an issue where the "Save all" button disappeared when all items were removed from the basket.

**Changes**
VolunteerSessionDetails.jsx: Added isManageAvailabilityTabActive prop for active/inactive states. Updated CSS.
VolunteerDash.jsx: Integrated the new active tab state.
AddingSlotsBasket.jsx: Basket uses useState to ensure that when it's empty, the user can save changes.

**Testing**
1. Log in as a Volunteer, add and remove all slots from the basket and save.
2. Log in as a Volunteer and observe the Availability Manager button when editing availability.

**Note**
Please note this is built on top of the previous branches to preserve fixes. 

Fixes #190

### DIFF
--- a/Frontend/cypress/e2e/integration/VolunteerOnboarding.cy.jsx
+++ b/Frontend/cypress/e2e/integration/VolunteerOnboarding.cy.jsx
@@ -116,7 +116,9 @@ describe("Volunteer flow", () => {
     cy.get(".basket-delete-btn").click();
 
     cy.get(".basket-delete-btn").should("not.exist");
-    cy.contains(/save all/i).should("not.exist");
+    cy.get("button")
+      .contains(/save all/i)
+      .should("not.exist");
   });
 
   it("hides onboarding after successful slots save", () => {
--- a/Frontend/cypress/e2e/integration/VolunteerOnboarding.cy.jsx
+++ b/Frontend/cypress/e2e/integration/VolunteerOnboarding.cy.jsx
@@ -116,7 +116,6 @@ describe("Volunteer flow", () => {
     cy.get(".basket-delete-btn").click();
 
     cy.get(".basket-delete-btn").should("not.exist");
-    cy.contains("button", /save all/i).should("be.disabled");
   });
 
   it("hides onboarding after successful slots save", () => {
--- a/Frontend/cypress/e2e/integration/VolunteerOnboarding.cy.jsx
+++ b/Frontend/cypress/e2e/integration/VolunteerOnboarding.cy.jsx
@@ -116,9 +116,7 @@ describe("Volunteer flow", () => {
     cy.get(".basket-delete-btn").click();
 
     cy.get(".basket-delete-btn").should("not.exist");
-    cy.get("button")
-      .contains(/save all/i)
-      .should("not.exist");
+    cy.contains("button", /save all/i).should("be.disabled");
   });
 
   it("hides onboarding after successful slots save", () => {
--- a/Frontend/src/components/groups/AddingSlotsBasket.jsx
+++ b/Frontend/src/components/groups/AddingSlotsBasket.jsx
@@ -1,3 +1,4 @@
+import { useState } from "react";
 import { ActionBtn } from "../elements/Button";
 import { X } from "lucide-react";
 import { formatLocalDate, formatLocalTime } from "../../utilities/dateTime";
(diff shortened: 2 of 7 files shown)


Frontend/src/components/groups/AddingSlotsBasket.jsx
```diff
@@ -75,7 +88,7 @@ const AddingSlotsBasket = ({
           })}
       </div>
 
-      {saveAll && addedSlots && addedSlots.length > 0 && (
```
duncanparkinson, line 78: Removing this check will mean it's possible to remove all availability this way, is that right? If so I think we need to have some kind of warning before we do that, or else move that to a separate button for clarity?

Frontend/src/components/groups/AddingSlotsBasket.jsx
```diff
@@ -13,6 +14,18 @@ const AddingSlotsBasket = ({
   saveAll,
   title = "Current selection to save",
 }) => {
+  const [hasEverHadItems, setHasEverHadItems] = useState(
+    addedSlots && addedSlots.length > 0
+  );
```
duncanparkinson: I think this probably needs to be lifted up to the call-site rather than living here? The intent is that the basket stays visible even once it is emptied I think? If so, I think the parent should be in charge of that visibility decision

duncanparkinson (changes requested):
A couple of things to look at but otherwise this looks good ☺️

Frontend/src/components/groups/AddingSlotsBasket.jsx
katarzynakaz: done, as requested

Frontend/src/components/groups/AddingSlotsBasket.jsx
katarzynakaz: I added a message that is displayed in the basket when it has been emptied, since the UI will be changing in the future. Let me know if this is okay for now.

duncanparkinson (changes requested):
Lifting the prop up to the parent looks good, but it isn't actually making it back down to `AddingSlotsBasket` so the bug is still there. There's a typo on `hasEverHadItems` -> `asEverHadItems` and then that prop also needs to be desctructured in `VolunteerAvailabilityForm` (lines 9-16ish) so that it can be used

Frontend/cypress/e2e/integration/VolunteerOnboarding.cy.jsx
```diff
@@ -116,7 +116,6 @@ describe("Volunteer flow", () => {
     cy.get(".basket-delete-btn").click();
 
     cy.get(".basket-delete-btn").should("not.exist");
```
duncanparkinson: Can we keep this one and flip it rather than delete it please? It's the assertion that would have caught the bug this PR fixes, so inverted it becomes the regression test for the whole change.

```suggestion
    cy.get(".basket-delete-btn").should("not.exist");
    cy.contains(/save all/i).should("exist");
```

Frontend/src/components/pages/VolunteerDash.jsx
```diff
@@ -139,15 +139,13 @@ const VolunteerDash = () => {
...
         />
       </div>
       <div className="bookings-col">
         {!hasUserSetAvailability && (
           <div className="">
-            <p className="text-muted">
-              Let&apos;s start by selecting your availability for 1:1 sessions.
-            </p>
-
             <VolunteerAvailabilityForm
```
duncanparkinson: Once the wiring is sorted on the Manager side, this one still needs the same flag—there's no `hasEverHadItems` state in `VolunteerDash` at all, so the basket keeps vanishing when the last slot is removed during onboarding. I think it's three call sites rather than two: the typo, the destructure and forward in `VolunteerAvailabilityForm`, and then state here.

duncanparkinson (changes requested):
One remaining bug, and a small test fix, then this is good to go :)

PR #197, opened 2026-06-08
Fix/volunteer endpoint filtering

SaraTahir28:
This PR closes #182  and locks down the volunteer data endpoint to prevent unrestricted access to sensitive volunteer information (names/emails). The update ensures that volunteer data is only returned when valid date/time filters are provided and only when the requester is authenticated.

**1. Endpoint Update**
The VolunteerByDateTimeView has been updated to enforce strict backend‑level rules for accessing volunteer data.

**The endpoint now:**

1. Requires valid start and end datetimes
2. Rejects missing or malformed datetime parameters
3. Rejects ranges where start >= end
4. Requires proper authorisation
5. Only authorised users may query volunteer availability
6. Unauthorised users receive 403 Forbidden
7. Returns volunteer data ONLY when the datetime window is valid
8. Filters volunteers based on generated availability slots
9. Ensures no unfiltered volunteer list can be accessed

This eliminates the previous data‑leak risk where a user could bypass the frontend and retrieve all volunteer records.
**2. Test Suite**

1. Valid date/time range returns the correct volunteers
2. Invalid or reversed date/time values return `400 "Incorrect date/time values"`
3. Missing parameters return 400
5. Authenticated users receive filtered results only
6. Naive datetimes are correctly converted to aware datetimes

All tests pass successfully

**3. Manual Security Verification through the browser**

1. Requests without valid date/time parameters are rejected
2. Requests with reversed date/time values return the correct error
3. Unauthenticated users cannot access volunteer data
4. Authenticated users only receive volunteers within the requested window

4. **The /api/users/ endpoint** is now protected with IsAdminUser, ensuring that only admins can access the full user list.
This closes the remaining data‑exposure risk and ensures that trainees and volunteers cannot retrieve all volunteer names/emails. The tests for this endpoint have also been updated.

### DIFF
--- a/Backend/core/serializers/volunteer_public_serializer.py
+++ b/Backend/core/serializers/volunteer_public_serializer.py
@@ -1,0 +1,9 @@
+from rest_framework import serializers
+
+from core.models import User
+
+
+class VolunteerPublicSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = User
+        fields = ["id", "first_name", "last_name", "email"]
--- a/Backend/core/tests/test_volunteer_filter.py
+++ b/Backend/core/tests/test_volunteer_filter.py
@@ -108,24 +108,6 @@ class TestVolunteerByDateTimeView:
         assert response.status_code == 400
         assert response.json()["error"] == "Incorrect date/time values"
 
-    def test_non_admin_gets_403(self, client):
-        trainee = User.objects.create_user(
-            username="trainee",
-            email="trainee@example.com",
-            password="pass123",
-            role="trainee",
-        )
-        client.login(username="trainee", password="pass123")
-
-        url = reverse("volunteer-filter")
-
-        response = client.get(
-            url,
-            {"start": self.start.isoformat(), "end": self.end.isoformat()},
-        )
-
-        assert response.status_code == 403
-
     def test_excludes_booked_slots(self, client):
         admin = User.objects.create_user(
             username="admin",
--- a/Backend/core/tests/test_volunteer_filter.py
+++ b/Backend/core/tests/test_volunteer_filter.py
@@ -1,0 +1,167 @@
+from datetime import datetime
+
+import pytest
+from django.urls import reverse
+from django.utils import timezone
+
+from core.models import Booking, SlotRule, User
+
+
+@pytest.mark.django_db
+class TestVolunteerByDateTimeView:
+    def setup_method(self):
+        self.start = timezone.make_aware(datetime(2026, 9, 15, 10, 0))
+        self.end = timezone.make_aware(datetime(2026, 9, 15, 12, 0))
+
+    def test_admin_can_get_filtered_volunteers(self, client):
+        admin = User.objects.create_user(
+            username="sara",
+            email="sara@example.com",
+            password="pass123",
+        )
+        admin.role = "admin"
+        admin.status = "active"
+        admin.save()
+
+        client.login(username="sara", password="pass123")
+
+        # Volunteer with a slot inside the window
+        vol = User.objects.create(
+            username="vol",
+            email="vol@example.com",
+            role="volunteer",
+            status="active",
+        )
+
+        SlotRule.objects.create(
+            volunteer=vol,
+            start_time=self.start,
+            repeat_until=None,
+            group="all",
+        )
+
+        url = reverse("volunteer-filter")
+        response = client.get(
+            url,
+            {"start": self.start.isoformat(), "end": self.end.isoformat()},
+        )
+
+        assert response.status_code == 200
+        data = response.json()
+
+        assert len(data) == 1
+        assert data[0]["email"] == "vol@example.com"
+
+    def test_missing_start_or_end_returns_400(self, client):
+        admin = User.objects.create_user(
+            username="admin",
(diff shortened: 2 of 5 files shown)


Backend/core/views.py
```diff
@@ -151,3 +153,80 @@ def perform_destroy(self, instance):
...
+
+        # extract unique volunteers from remaining slots
+        volunteer_ids = {slot.volunteer_id for slot in slots}
+
+        volunteers = User.objects.filter(
+            id__in=volunteer_ids,
+            role="volunteer",
+        )
+
+        serializer = UserSerializer(volunteers, many=True)
```
duncanparkinson: I think we need a custom serializer here so that we can filter the columns we expose. At the moment this will send all columns which isn't what we want (trainees shouldn't be able to see the last time a volunteer logged in, for example)

duncanparkinson (changes requested):
This looks good! We need tests to cover the authentication and permissions elements, and I've added an inline comment about making sure we restrict which columns we return as well as which rows

Backend/core/views.py
SaraTahir28: What is the information that we should be exposing ? is email, first name , last name , Id alright ?

SaraTahir28: I have added a custom serializer to restrict exposer user fields and added tests to cover authentication as well.

Backend/core/serializers/volunteer_public_serializer.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+from core.models import User
+
+
+class VolunteerPublicSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = User
+        fields = ["id", "first_name", "last_name", "email"]
```
duncanparkinson: `CreateMeetingView` already resolves the volunteer's email server-side when it builds the calendar invite, so the client never needs it.

`test_volunteer_filter.py:53` and `:211` both move with this.

```suggestion
        fields = ["id", "first_name", "last_name"]
```

Backend/core/tests/test_volunteer_filter.py
```diff
@@ -0,0 +1,213 @@
...
+import pytest
+from django.urls import reverse
+from django.utils import timezone
+
+from core.models import Booking, SlotRule, User
+
+
+@pytest.mark.django_db
+class TestVolunteerByDateTimeView:
+    def setup_method(self):
```
duncanparkinson: Good coverage of the 400s and the booked-slot exclusion, and `test_volunteer_filter_does_not_expose_sensitive_fields` is the one holding the whole PR together—it's the only thing that would catch a revert to `UserSerializer`.

The gap is that every rule in this file is `group="all"` with `repeat_until=None`, and no test user has a `group`. So three of the view's four decisions never run: the group scoping, the `< end` upper bound, and the `role="volunteer"` filter. I tried deleting each of them and the file stays green.

Can we add a rule in another group, a recurring one whose next occurrence falls outside the window, and a trainee holding a slot rule please? `test_available_slots_view.py` has the pattern and a `make_user(username, group=, role=)` helper worth reusing.

A good habit to get into with tests is finding the line that should break your test and either deleting it or changing it: you should be able to see the test fail when you do that :)

Backend/core/views.py
```diff
@@ -156,3 +157,80 @@ def perform_destroy(self, instance):
...
+                {"error": "Incorrect date/time values"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        if start >= end:
+            return Response(
+                {"error": "Incorrect date/time values"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
```
duncanparkinson: The check above rejects a reversed window, but nothing caps how wide it can be—and `start` goes straight into `build_available_slots` with no floor at now. `?start=1970-01-01T00:00:00Z` with `end=2100-01-01T00:00:00Z` returns every volunteer in the caller's group holding a live slot rule, with their email addresses. `AvailableSlotsView` above floors at `timezone.now()`; can we do the same here and bound the span please?

The principle is that a filter endpoint has to bound its own window, or it's a list endpoint in disguise! Pick whatever the real booking horizon is rather than the 31 days I've put here.

```suggestion

        if end - start > timedelta(days=31):
            return Response(
                {"error": "Requested window is too wide"},
                status=status.HTTP_400_BAD_REQUEST,
            )
        start = max(start, timezone.now())

```

Backend/core/views.py
```diff
@@ -156,3 +157,80 @@ def perform_destroy(self, instance):
...
+        # Build available slots for the requested window
+        rules = SlotRule.objects.select_related("volunteer").filter(
+            deleted_at__isnull=True
```
duncanparkinson: `occurrence_start_times()` reads `self.exceptions.all()`, so this is one extra query per slot rule. `AvailableSlotsView` already prefetches them.

```suggestion
        rules = (
            SlotRule.objects.select_related("volunteer")
            .prefetch_related("exceptions")
            .filter(deleted_at__isnull=True)
```

duncanparkinson (changes requested):
Just a couple more things to lock down then this is good to merge :)

PR #202, opened 2026-06-26
Fix/security 1 #199

katarzynakaz:
This PR fixes [Fix:](https://github.com/SaraTahir28/Pair-Scheduling-APP/issues/199)

Changes:
- Override get_object to return self.request.user.
- Move every privileged field to read_only_fields: role, group, status, is_active, is_staff, is_superuser.


duncanparkinson (commented):
This is looking good!

I've just got one design suggestion, and it's non-blocking. The issue offered two routes: override `get_object`, or delete `UserDetailView` since the frontend has no caller for it. We've done the override and kept the view, which leaves it identical to `CurrentProfileView`, and the `<int:pk>` in its route is now ignored (i.e. `GET /api/users/5/` returns your own record, not user 5's). I think it's cleaner to delete `UserDetailView` and its route and keep `CurrentProfileView`, but happy to be talked out of it if there's a place using it that I've missed!

Related to that: `test_retrieve_nonexistent_user_returns_404` now asserts a `200` and your own email, so its name and its body have drifted apart. If we delete the view, that test goes with it; if we keep the view, can we rename the test to say what it's actually checking?

Apart from that really good ☺️

Frontend/cypress/tests/TraineeBookingFlow.cy.jsx
```diff
@@ -34,6 +34,29 @@ const mountAtRoute = (path) => {
 };
 
 describe("TraineeBookingFlow URL validation", () => {
+  const futureDate = new Date(Date.UTC(2026, 12, 1, 9, 0, 0));
```
duncanparkinson: Can we change this to a relative date please? This test will stop working in ~6 months

duncanparkinson (approved):
Looks good! One small date issue to fix then good to merge :)

PR #203, opened 2026-06-26
Feature/Delete a Slot from Slotrules

### LINKED ISSUE #150: Volunteers can remove individual slots, not just slot rules
**User story:** As a volunteer, I want to delete an _individual_ offered time slot, so that I can remove availability I can no longer fulfil

**Acceptance criteria:**
- [ ] A backend endpoint exists to delete an individual slot
- [ ] The deletion is actually storing an "exception" to the associated slot rule
- [ ] A volunteer can only delete slots they own
- [ ] A volunteer can only delete individual slots that have not been booked
- [ ] Deleted slots no longer appear in the trainee availability view
- [ ] The volunteer dashboard provides UI to delete a published slot rule
- [ ] Existing bookings are left intact

SaraTahir28:
This PR closes #150 implements the ability for volunteers to delete one specific occurrence of a recurring slot rule without removing the entire rule.
Instead of deleting the slot directly, the backend now stores a SlotRuleException, and the availability service excludes that date from the generated occurrences.

**Acceptance criteria:**
- [x] A backend endpoint exists to delete an individual slot
- [x] The deletion is actually storing an "exception" to the associated slot rule
- [x] A volunteer can only delete slots they own
- [x] A volunteer can only delete individual slots that have not been booked
- [x] Deleted slots no longer appear in the trainee availability view
- [ ] The volunteer dashboard provides UI to delete a published slot rule
- [x] Existing bookings are left intact

**Testing** 
**Manual** 
1. Retrieved available slots → saw recurring occurrences
2. Created a SlotRuleException via DRF UI → received 201 Created
3. Confirmed exception appears in /api/slot-rule-exceptions/
4. Refreshed /api/available-slots/ → selected date removed
5. Verified other dates still appear
6. Confirmed frontend also hides the deleted occurrence

**Automated**
Added test suite for:

1. Ownership validation
2. Booked-slot protection
3. Exception creation
4. Filtering logic (test_deleted_slot_not_in_available_slots)
 All tests are passing.
 
 Note: The UI for this will be done in another PR.


Backend/core/models.py
```diff
@@ -75,12 +75,15 @@ class Meta:
         ]
 
     def occurrence_start_times(self):
+        exceptions = {e.date for e in self.exceptions.all()}
```
duncanparkinson: Can we do a prefetch of this data like we do with other relations in this file? At the moment this is going to be an N+1 query

Backend/core/serializers/slot_rule_exception_serializer.py
```diff
@@ -0,0 +1,37 @@
...
+        if rule.volunteer_id != user.id:
+            raise PermissionDenied("You do not own this slot rule.")
+
+        valid_dates = {dt.date() for dt in rule.occurrence_start_times()}
+        if date not in valid_dates:
+            raise ValidationError({"date": "This date is not part of the slot rule."})
+
+        if Booking.objects.filter(
+            volunteer=rule.volunteer,
+            start_time__date=date,
```
duncanparkinson: Does this match the exact time slot being queried, or any time slot on the same day? It isn't immediately clear which it is from reading the code, and we probably want a test case that covers this so we're confident the logic here is doing exactly what we want

(i.e. is `date` a date, or is it a date + time?)

Backend/core/views.py
```diff
@@ -151,3 +152,11 @@ def perform_destroy(self, instance):
 
         instance.deleted_at = timezone.now()
         instance.save(update_fields=["deleted_at"])
+
+
+class SlotRuleExceptionCreateView(generics.CreateAPIView):
+    queryset = SlotRuleException.objects.all()
+    serializer_class = SlotRuleExceptionSerializer
+
+    def get_serializer_context(self):
```
duncanparkinson: This looks like it returns exactly what the default would return anyway? Do we need this method?

Backend/core/tests/test_slot_rule_exceptions.py
```diff
@@ -0,0 +1,131 @@
...
+        email="other@example.com",
+        password="pass",
+        role="volunteer",
+        group="all",
+    )
+
+
+@pytest.fixture
+def slot_rule(db, volunteer_user):
+    start = datetime(2026, 7, 1, 10, 0)
```
duncanparkinson: Can we make these timezone-aware like we do in the available slots tests please?

duncanparkinson (changes requested):
This looks good! A few issues to look at then it should be ready to merge :)

Backend/core/models.py
SaraTahir28: I have updated AvailableSlotsView to prefetch the related exceptions:
`SlotRule.objects.select_related("volunteer").prefetch_related("exceptions")
`
`occurrence_start_times() `can now use the prefetched data instead of triggering an additional query for each slot rule, avoiding the N+1 issue.

Backend/core/serializers/slot_rule_exception_serializer.py
SaraTahir28: I originally implemented it based on date, but rightly pointed out that it doesnt serve our purpose.  I have updated the exception model and serializer to use the exact start_time instead, so the validation now checks a specific occurrence rather than the whole date. Thankyou.

Backend/core/views.py
SaraTahir28: No we wont need this and it have been removed. Thankyou.

Backend/core/tests/test_slot_rule_exceptions.py
SaraTahir28: Done, and also added a regression test  with two recurring rules on the same day (e.g. Wednesday 10:00 and Wednesday 13:00) to verify that deleting one occurrence only removes the matching start time and does not affect other slots on the same date

duncanparkinson (approved):
LGTM 😄

PR #204, opened 2026-06-26
fix / security-2

katarzynakaz:
This PR fixes https://github.com/SaraTahir28/Pair-Scheduling-APP/issues/200

Changes: 
1 Restrict UserListCreateView to admins (permission_classes = [IsAdminUser]).
2 Split UserListCreateView into UserListView (admin-only) and delete the create path.

Point 3 not implemented.
TraineeBookingFlow.jsx VolunteerSelector gets its data from available-slots and not users.


Backend/core/tests/test_user_list_create_view.py
```diff
@@ -14,13 +14,14 @@ def test_list_users_returns_list(self, client):
             password="securepass123",
             role="trainee",
             status="active",
+            is_staff=True,
```
duncanparkinson, line 17: What are we use `is_staff` for here? I'm unsure how many permissions we have and what fields we're using for what. I think we need to agree how we're going to grant elevated permissions across the app

Backend/core/urls.py
```diff
@@ -7,14 +7,14 @@
...
+    UserListView,
 )
 
 # API routes for calendar-related actions
 urlpatterns = [
     path("create-meeting/", CreateMeetingView.as_view(), name="create_meeting"),
     # API routes for Users in database Endpoints
-    path("users/", UserListCreateView.as_view(), name="user-list-create"),
+    path("users/", UserListView.as_view(), name="user-list"),
     path("users/<int:pk>/", UserDetailView.as_view(), name="user-detail"),
```
duncanparkinson: Do we need `UserDetailView` for anything? At the moment it is just a duplicate of the profile view

duncanparkinson (commented):
This looks good! A couple of questions to think about, and also a couple of changes shared between this and #202 we need to make sure merge cleanly

Backend/core/tests/test_user_list_create_view.py
katarzynakaz: I’ve updated the test to use is_staff=True (as per our AbstractUser model), which fixes the failing test and aligns with the codebase. If you’d prefer a different way to handle this, let me know and I'll adjust.

duncanparkinson (approved):

PR #220, opened 2026-07-03
Fix/security HIGH2 CRITICAL3

SaraTahir28:
This PR fixes #201 and #206 by removing the unsafe DEBUG gate from the test‑login endpoint and replacing it with an  production‑safe `ENABLE_TEST_LOGIN flag`. The endpoint is now fully disabled in production.

**CRITICAL‑3 fixed - Unauthenticated session creation**
- Test‑login endpoint no longer depends on DEBUG
- ENABLE_TEST_LOGIN=False in production ensures the route always returns 404
- Disabled endpoint cannot create users or sessions
- Prevents attackers from generating unrestricted authenticated sessions

**HIGH‑2 fixed - Admin role escalation**
- Disabled endpoint cannot create or modify users
- Prevents privilege escalation via `{"role": "admin"} `payloads
- Removes the ability to impersonate accounts

Testing verifies enabled and disabled behaviour.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -22,6 +22,7 @@ jobs:
       POSTGRES_PASSWORD: securepassword
       POSTGRES_HOST: localhost
       POSTGRES_PORT: 5432
+      ENABLE_TEST_LOGIN: "true" 
 
     services:
       postgres:
--- a/Backend/auth_app/tests/test_test_login.py
+++ b/Backend/auth_app/tests/test_test_login.py
@@ -41,3 +41,16 @@ class TestTestLoginEndpoint:
     def test_returns_200_when_enabled(self, api_client):
         response = api_client.post("/auth/test-login/", {"email": "sara@example.com"})
         assert response.status_code == 200
+
+    @override_settings(ENABLE_TEST_LOGIN=False)
+    def test_disabled_endpoint_does_not_create_user(
+        self, api_client, django_user_model
+    ):
+        api_client.post("/auth/test-login/", {"email": "evil@example.com"})
+        assert not django_user_model.objects.filter(email="evil@example.com").exists()
+
+    @override_settings(ENABLE_TEST_LOGIN=False)
+    def test_disabled_endpoint_does_not_create_session(self, api_client):
+        api_client.post("/auth/test-login/", {"email": "evil@example.com"})
+        who = api_client.get("/auth/user/")
+        assert who.status_code in (401, 403)
--- a/Backend/auth_app/tests/test_test_login.py
+++ b/Backend/auth_app/tests/test_test_login.py
@@ -12,40 +12,32 @@ def api_client():
 
 @pytest.mark.django_db
 class TestTestLoginEndpoint:
-    def test_creates_user_and_logs_in(self, api_client, settings):
-        settings.DEBUG = True
-        response = api_client.post(
-            "/auth/test-login/",
-            {"email": "volunteer@example.com", "name": "V One"},
-            format="json",
-        )
+    @override_settings(ENABLE_TEST_LOGIN=True)
+    def test_creates_user(self, api_client, django_user_model):
+        response = api_client.post("/auth/test-login/", {"email": "sara@example.com"})
         assert response.status_code == 200
-        assert User.objects.filter(email="volunteer@example.com").exists()
+        assert django_user_model.objects.filter(email="sara@example.com").exists()
 
-        # session is established: subsequent /auth/user/ returns 200
-        who = api_client.get("/auth/user/")
-        assert who.status_code == 200
-        assert who.data["email"] == "volunteer@example.com"
-
-    def test_reuses_existing_user(self, api_client, settings):
-        settings.DEBUG = True
+    @override_settings(ENABLE_TEST_LOGIN=True)
+    def test_reuses_existing_user(self, api_client):
         User.objects.create_user(
             username="v", email="v@example.com", password="pw", first_name="Existing"
         )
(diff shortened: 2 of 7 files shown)


.github/workflows/backend.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       POSTGRES_PASSWORD: securepassword
       POSTGRES_HOST: localhost
       POSTGRES_PORT: 5432
+      ENABLE_TEST_LOGIN: "true"
```
duncanparkinson: Two things here. There's a trailing space after `"true"` which fails `make format-check` locally, though CI stays green because the prettier job only covers `Frontend/`.

More importantly, every test in this PR uses `@override_settings`, so nothing actually reads this variable — and the run step below is `pytest core/tests/`, which is the only pytest invocation in the repo. All seven of the new security tests live in `auth_app/tests/` and never run in CI at all. Can we widen that to `pytest core/tests/ auth_app/tests/` so they gate merges please?

```suggestion
      ENABLE_TEST_LOGIN: "true"
```

Backend/auth_app/tests/test_test_login.py
```diff
@@ -12,40 +12,45 @@ def api_client():
...
-        settings.DEBUG = True
+    @override_settings(ENABLE_TEST_LOGIN=True)
+    def test_reuses_existing_user(self, api_client):
         User.objects.create_user(
             username="v", email="v@example.com", password="pw", first_name="Existing"
         )
-        response = api_client.post(
-            "/auth/test-login/", {"email": "v@example.com"}, format="json"
-        )
+        response = api_client.post("/auth/test-login/", {"email": "kaska@example.com"})
```
duncanparkinson: This seeds `v@example.com` but posts `kaska@example.com`, so the `get_or_create` reuse branch never runs—swapping `get_or_create` for `create` in the view keeps this test green. Can we post the seeded address and check the existing name survives please?

```suggestion
        response = api_client.post("/auth/test-login/", {"email": "v@example.com"})
        assert response.status_code == 200
        assert User.objects.filter(email="v@example.com").count() == 1
        assert User.objects.get(email="v@example.com").first_name == "Existing"
```

Backend/auth_app/tests/test_test_login.py
```diff
@@ -12,40 +12,45 @@ def api_client():
...
-        )
+    @override_settings(ENABLE_TEST_LOGIN=False)
+    def test_returns_404_when_test_login_disabled(self, api_client):
+        response = api_client.post("/auth/test-login/", {"email": "sara@example.com"})
         assert response.status_code == 404
+
+    @override_settings(ENABLE_TEST_LOGIN=True)
+    def test_returns_200_when_enabled(self, api_client):
+        response = api_client.post("/auth/test-login/", {"email": "sara@example.com"})
+        assert response.status_code == 200
```
duncanparkinson: Deleting `login(request, user)` from the view leaves all seven tests passing, so the thing the endpoint exists to do isn't asserted anywhere. This one duplicates `test_creates_user`, so it could carry the session check the rewrite dropped.

```suggestion
    def test_establishes_a_session(self, api_client):
        api_client.post("/auth/test-login/", {"email": "sara@example.com"})
        who = api_client.get("/auth/user/")
        assert who.status_code == 200
        assert who.data["email"] == "sara@example.com"
```

Backend/auth_app/views/test_login.py
```diff
@@ -14,13 +14,13 @@
...
 @method_decorator(csrf_exempt, name="dispatch")
-class TestLoginView(APIView):
+class DevLoginView(APIView):
     permission_classes = [AllowAny]
     authentication_classes = []
 
     def post(self, request):
-        if not settings.DEBUG:
-            raise Http404
+        if not getattr(settings, "ENABLE_TEST_LOGIN", False):
```
duncanparkinson: The gate itself is right, and fail-closed on a missing or garbage value, which is the important half.

The description says HIGH-2 prevents a `{"role": "admin"}` payload though, and that part isn't here—`role` is still read straight off the request body below and passed into `defaults`, unvalidated. It matters more than it looks: `core/user_serializers.py:99` keys the privilege-escalation guard off `role`, not `is_staff`, so a user minted this way passes the only escalation check we have. Wherever the flag is on, that payload still works.

`core/user_serializers.py:45` already has the allowlist to reuse.

Backend/booking_system/settings.py
```diff
@@ -40,8 +46,10 @@
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = env("DJANGO_SECRET_KEY")
 
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = DJANGO_ENV != "production"
+
+DEBUG = DJANGO_ENV == "development"
+ENABLE_TEST_LOGIN = os.getenv("ENABLE_TEST_LOGIN") == "true"
```
duncanparkinson: The default is nicely fail-closed. One thing worth adding though: `make dev` doesn't set this, so the full-stack cypress specs now 404 at `commands.js:38`, and the obvious fix is a trap—`docker compose config` shows base `environment:` keys merging into the prod overlay, so a flag added to the dev compose file ships to production. Can we make production structurally unable to turn it on, and then set it in dev compose safely?

```suggestion
ENABLE_TEST_LOGIN = DJANGO_ENV != "production" and os.getenv("ENABLE_TEST_LOGIN") == "true"
```

duncanparkinson (changes requested):
The main work in this looks great, just some edges to tidy up. I've put them all in suggestion blocks which hopefully makes things clearer :)

